PR #6, opened 2018-04-13
RDM-2012 Add dependency check

adr1ancho:

RDM-2012

### Change description ###
Add dependency check

```
[x] Yes
[ ] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ plugins {
     id 'application'
     id 'io.spring.dependency-management' version '1.0.4.RELEASE'
     id 'org.springframework.boot' version '1.5.6.RELEASE'
-    id 'org.owasp.dependencycheck' version '3.1.1'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -414,3 +414,6 @@ task codeCoverageReport(type: JacocoReport, dependsOn: allprojects*.test) {
         xml.destination = file("${buildDir}/reports/jacoco/jacocoTestReport.xml")
     }
 }
+
+// for cnp pipeline
+task dependencyCheck(dependsOn: dependencyCheckAnalyze)
```
Daniel-Lam: Where is `dependencyCheckAnalyze`? Is it supposed to be another task?

adr1ancho: New pipeline should be dependencyCheckAnalyze
Old pipeline was dependencyCheck
And there can be more than one pipeline out there

Daniel-Lam: OK, that's fine

PR #14, opened 2018-05-04
RDM 2053 - Custom Button Label for Events

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-2053

The button text on the submit page to be made configurable for each service and each event.
For this purpose a new column named `EndButtonLabel` is added to the `CaseEvent` tab of the definition excel file.
If the value has been omitted in the definition then set it to 'Submit'

```
[ ] Yes
[X] No
```


application/src/test/java/uk/gov/hmcts/net/ccd/definition/store/BaseTest.java
```diff
@@ -49,7 +49,7 @@
 })
 @TestPropertySource(locations = "classpath:test.properties")
 public abstract class BaseTest {
-    public static final String EXCEL_FILE_CCD_DEFINITION = "/CCD_TestDefinition_V29_RDM-1831.xlsx";
+    public static final String EXCEL_FILE_CCD_DEFINITION = "/CCD_TestDefinition_V30.xlsx";
```
mario-paniccia: how come we're not adding anymore the story number to the file name? that proved useful to avoid collisions when multiple persons working on it at the same time

PR #15, opened 2018-05-08
Update s2s api

timja-kainos:
### JIRA link (if applicable) ###
?

### Change description ###
Update S2S api version

Based on https://github.com/hmcts/ccd-data-store-api/pull/15/files

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

PR #17, opened 2018-05-08
Update dependencies

timja-kainos:
HikariCP wasn't working in docker without updating deps, looked at it with V and Daniel


timja-kainos:
Moved into #15 as builds are very flakey atm and only want to do one

PR #18, opened 2018-05-09
RDM-2055 Add showSummaryContentOption column to CaseEventToFields tab…

andrewfolga:
…. Make sure it is returned in GET /data/case-type/{ctid} call.

https://tools.hmcts.net/jira/browse/RDM-2055

### Change description ###
Read only summary fields
Configuring Event To Field for readonly element on submit page
https://tools.hmcts.net/confluence/display/RCCD/Configurable+ReadOnly+Display+of+Fields+at+the+end+of+an+Event

```
[ ] Yes
[x] No
```


application/src/test/java/uk/gov/hmcts/net/ccd/definition/store/rest/MultipleControllersEndpointIT.java
```diff
@@ -35,6 +35,38 @@
     private static final String ROLES_URL = "/api/user-roles/%s";
     private static final String WORKBASKET_INPUT_DEFINITION_URL = "/api/display/work-basket-input-definition/%s";
     private static final String JURISDICTIONS_URL = "/api/data/jurisdictions";
+    private static final String CASE_TYPE_URL = "/api/data/case-type/%s";
+
+    @Test
+    public void shouldReturnCaseType() throws Exception {
```
mario-paniccia: Don't we have a IT test already that retrieves a CaseType that we can update instead of adding a brand new? Asking because IT tests are slow

andrewfolga: I checked and we do not have a full e2e test for get CaseType. There is a class CaseDefinitionControllerTest that is using MockMvc but is also using mocks so is not truly e2e.

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/model/CaseEventField.java
```diff
@@ -52,7 +53,8 @@ public void setShowCondition(String showCondition) {
     }
 
     /**
-     */
+     * whether field is shown with the change option
```
andrewfolga: Check and rename if needed

andrewfolga: This is ok turns out.

PR #20, opened 2018-05-10
Remove unneeded options

timja-kainos:
### Change description ###
Only deploy to saat or sprod, nayab was accidentally deploying to sandbox before

```
[ ] Yes
[ ] No
```


vlaurin:
Removing the usernames/password is okay for now as we are wrongly using IDAM test endpoint; but this will need to be re-instated with the proper authentication flow so that we can stop relying on IDAM test endpoint.

PR #38, opened 2018-05-31
RDM-2259: Fix wrong storing of integers in definition as decimals

mario-paniccia:
https://tools.hmcts.net/jira/browse/RDM-2259
This also fixes:
https://tools.hmcts.net/jira/browse/RDM-2260

NOTE to QA: This story changes the way we store min/max values on the DB. we need to verify the application can still read the old definitions. 

http://poi.apache.org/apidocs/org/apache/poi/ss/usermodel/DataFormatter.html#formatCellValue-org.apache.poi.ss.usermodel.Cell-

```
[ ] Yes
[ X] No
```


excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/parser/SpreadsheetParser.java
```diff
@@ -22,11 +22,12 @@
     private static final Logger logger = LoggerFactory.getLogger(SpreadsheetParser.class);
```
mario-paniccia: I wanted to add a test for this, but this class at the moment hasn't got any and it's quite untestable as it is. I would need to refactor it to allow testability. If you think we need to do this I'm going to do that.

PR #55, opened 2018-06-17
Add explicit declaration of environment variable for authorised services

Daniel-Lam:
Declare `DEFINITION_STORE_S2S_AUTHORISED_SERVICES` environment variable explicitly, as this is required on the Azure platform (which does not use the default values defined in `application.properties`).

None.

### Change description ###
Fixes a problem in the CNP `Demo` environment, where Case Definition Store was refusing access by Case Admin Web application because no authorised services were defined (due to the missing environment variable).

This change has already been tested manually by adding the missing environment variable via the Azure Portal and confirming that Case Admin Web is now able to access Case Definition Store.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -68,6 +68,9 @@ module "case-definition-store-api" {
     IDAM_USER_URL = "${var.idam_api_url}"
     IDAM_S2S_URL = "${local.s2s_url}"
     DEFINITION_STORE_IDAM_KEY = "${data.vault_generic_secret.definition_store_item_key.data["value"]}"
+
+    DEFINITION_STORE_S2S_AUTHORISED_SERVICES = "${var.authorised-services}"
+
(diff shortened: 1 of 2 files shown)


infrastructure/main.tf
```diff
@@ -68,6 +68,9 @@ module "case-definition-store-api" {
     IDAM_USER_URL = "${var.idam_api_url}"
     IDAM_S2S_URL = "${local.s2s_url}"
     DEFINITION_STORE_IDAM_KEY = "${data.vault_generic_secret.definition_store_item_key.data["value"]}"
+
+    DEFINITION_STORE_S2S_AUTHORISED_SERVICES = "${var.authorised-services}"
+
     USER_PROFILE_HOST = "http://ccd-user-profile-api-${local.env_ase_url}"
```
adr1ancho: Nice when you put spaces around this very important line

PR #77, opened 2018-07-04
RDM-1240 search cases by metadata - DB changes

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-1240

### Change description ###
To support metadata fields, column case_type_id in table case_field is made nullable so that metadata fields can be stored as global fields common to all case types. A new column data_field_type is added to identify case field as either METADATA or CASEDATA. 

```
[ ] Yes
[x] No
```


repository/src/main/resources/db/changelog/db.changelog-rdm-1240.1.xml
```diff
@@ -0,0 +1,13 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="rdm-1240-1" author="giri.benadikar@hmcts.net">
+        <addColumn tableName="case_field">
+            <column name="data_field_type" type="varchar(10)" defaultValue="CASE_DATA">
```
vlaurin: I believe it would be nicer to create an enum type first, then type this column with the enum instead of `varchar(10)`

repository/src/main/resources/db/changelog/db.changelog-rdm-1240.3.xml
```diff
@@ -0,0 +1,83 @@
...
+
+    <changeSet id="rdm-1240-3" author="giri.benadikar@hmcts.net">
+        <insert tableName="case_field">
+            <column name="reference" value="jurisdiction"/>
+            <column name="live_from" value="${now}"/>
+            <column name="label" value="Jurisdiction"/>
+            <column name="hidden" value="false"/>
+            <column name="security_classification" value="PUBLIC"/>
+            <column name="field_type_id"
+                    valueComputed="(select id from field_type where reference = 'Text' and version = 1 and jurisdiction_id is null)"/>
```
vlaurin: It may not be relevant at the moment as we only have 1 version of base types, but a better approach would be to remove `and version = 1` and replace it by:
```
order by version desc limit 1
```
to make sure we are using the latest version of the global field type.

repository/src/main/resources/db/changelog/db.changelog-rdm-1240.3.xml
```diff
@@ -0,0 +1,83 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="rdm-1240-3" author="giri.benadikar@hmcts.net">
+        <insert tableName="case_field">
+            <column name="reference" value="jurisdiction"/>
```
vlaurin: I believe there was a reason to use the exact notation with upper case and square brackets as reference for metadata types:
```
[JURISDICTION]
```

vlaurin: This would remove the need for additional logic when mapping to the definition and it also would prevent collision with flexible case fields as [] are not valid characters for those (this is relevant for the state case field which is created as non-global).

vlaurin: This comment also applies to any references below

gbenadikar: The additional logic when mapping metadata fields to definition was trivial compared to sanitising the name of metadata fields for retrievals (the names are in-line with the standard search API).

repository/src/main/resources/db/changelog/db.changelog-rdm-1240.3.xml
gbenadikar: Done

repository/src/main/resources/db/changelog/db.changelog-rdm-1240.1.xml
gbenadikar: Done

repository/src/main/resources/db/changelog/db.changelog-rdm-1240.3.xml
vlaurin: Ok, then we have the risk of collision for state which means we need to add a validation rule around that in definition store to prevent anyone from using `state` as a field name.

gbenadikar:
This PR is merged into #79

PR #94, opened 2018-07-27
NEW - RDM-1240 - search cases by metadata

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-1240

### Change description ###
- Definition import process changed to add metadata fields to the parser context to support defining metadata fields in excel tabs.
- State is added as a dynamic metadata field during import as it depends on case type
- Introduced validations not to allow metadata fields in case field tab and should be read-only on case event fields tab
- When retrieving case type information, the metadata fields will be added to the case type - case fields with fixed list values if a field is a fixed list

```
[ ] Yes
[ x] No
```


gbenadikar:
separated PR

PR #112, opened 2018-08-15
RDM-2484 Remove copied keys

dwaynebailey:
Remove keys that we tried to copy into the Azure vault

https://tools.hmcts.net/jira/browse/RDM-2484 part 1

```
[ ] Yes
[X] No
```


infrastructure/main.tf
```diff
@@ -145,9 +145,3 @@ resource "azurerm_key_vault_secret" "gw_oauth2_secret" {
   value = "${data.vault_generic_secret.gateway_oauth2_client_secret.data["value"]}"
```
vlaurin: This one and the one above are copied from Hashi Vault as well; should they be kept?

vlaurin: 👍

PR #122, opened 2018-09-04
[PREVIEW] Use Azure vault

dwaynebailey:
https://tools.hmcts.net/jira/browse/RDM-2484

All the work to get the vault using ccd-$env exept removing the existing
local vault.  A few cleanup to check, where I've removed unused keys
(they're used in Jenkins but we won't need to use them in TF to copy
into the vault)

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -40,8 +40,9 @@ data "azurerm_key_vault" "ccd_shared_key_vault" {
   resource_group_name = "${local.sharedResourceGroup}"
 }
 
-data "vault_generic_secret" "definition_store_item_key" {
-  path = "secret/${var.vault_section}/ccidam/service-auth-provider/api/microservice-keys/ccd-definition"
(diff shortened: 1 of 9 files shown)


infrastructure/main.tf
```diff
@@ -32,19 +37,12 @@ locals {
...
+  resource_group_name = "${local.sharedResourceGroup}"
 }
 
-data "vault_generic_secret" "gateway_idam_key" {
-  path = "secret/${var.vault_section}/ccidam/service-auth-provider/api/microservice-keys/ccd-gw"
-}
-
-data "vault_generic_secret" "gateway_oauth2_client_secret" {
-  path = "secret/${var.vault_section}/ccidam/idam-api/oauth2/client-secrets/ccd-gateway"
+data "azurerm_key_vault_secret" "definition_store_item_key" {
```
vlaurin: It says `item` instead of `idam`, and actually, if it's s2s it's not idam anymore; so something closer to the actual secret name would be better I reckon.

PR #124, opened 2018-09-05
RDM-2707_RDM_2708 - First cut of Read User roles service implementation

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-2707

As a CMCA, I want to the ability to view a list of existing user roles in ccd for a selected/assigned jurisdiction backend implementation of which is done under
https://tools.hmcts.net/jira/browse/RDM-2708

```
[ ] Yes
[x] No
```

### DIFF
--- a/rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/UserRoleController.java
+++ b/rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/UserRoleController.java
@@ -67,4 +67,13 @@ class UserRoleController {
         @ApiParam(value = "Roles", required = true) @PathVariable("roles") List<String> roles) {
         return this.userRoleService.getRoles(roles);
     }
+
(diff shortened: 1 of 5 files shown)


rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/UserRoleController.java
```diff
@@ -67,4 +67,13 @@ UserRole userRoleGet(@RequestParam("role") @NotNull byte[] roleBase64EncodedByte
         @ApiParam(value = "Roles", required = true) @PathVariable("roles") List<String> roles) {
         return this.userRoleService.getRoles(roles);
     }
+
+    @RequestMapping(value = "/all-roles", method = RequestMethod.GET, produces = {"application/json"})
```
andrewfolga: Can we change the Mapping to `/user-roles` to make the naming consistent with other endpoints? I think we do not need to mention all explicitly so leaving it out would not obscure readability.

PR #129, opened 2018-09-11
RDM-2903 Make retrieval of Case Types faster in Create Case filters

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-2903

Added Case Event Lite object for returning the Events of a Case Type on GET /jurisdictions query. Also added pre_states to let the FE decide if the event is a "Create Event"

```
[ ] Yes
[X] No
```

### DIFF
--- a/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/EntityToResponseDTOMapper.java
+++ b/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/EntityToResponseDTOMapper.java
@@ -57,6 +57,14 @@ public interface EntityToResponseDTOMapper {
     )
     CaseEvent map(EventEntity eventEntity);
 
+    @Mapping(
+        expression = "java(eventLiteEntity.getPreStates().isEmpty() ? java.util.Collections.emptyList()" +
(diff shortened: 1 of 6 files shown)


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/EntityToResponseDTOMapper.java
```diff
@@ -57,6 +57,14 @@
     )
     CaseEvent map(EventEntity eventEntity);
 
+    @Mapping(
+        expression = "java(eventLiteEntity.getPreStates().isEmpty() ? java.util.Collections.emptyList()" +
```
andrewfolga: Aren't we missing this line from the mapping?:
`java(eventEntity.isCanCreate() ? java.util.Collections.emptyList()`
Seems that we have it for:
`    CaseEvent map(EventEntity eventEntity);`

tzarsmango: Thought it was not relevant at first but you're right, added the missing checks 👍

PR #186, opened 2018-10-15
RDM-3144 CCD displays 3 events all called ‘Apply for a divorce’ when logged in with solicitor account

tzarsmango:
added ACLs to CaseTypeLite/EventLite classes in `/jurisdictions` query. These ACLs will be used for CRUD filtering in `ccd-data-store-api` `/userProfile` query so that unauthorised Case Types, States or Events won't be visible to unauthorised users
+ removed StateLiteEntity and started usingStateEntity which is practically same weight

The relevant  [case-data-api PR #228](https://github.com/hmcts/ccd-data-store-api/pull/228) must be merged after this one is deployed...
```
[ ] Yes
[X] No
```


repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/CaseTypeLiteACLEntity.java
```diff
@@ -0,0 +1,23 @@
...
+import javax.validation.constraints.NotNull;
+import java.io.Serializable;
+
+@Table(name = "case_type_user_role")
+@Entity
+public class CaseTypeLiteACLEntity extends Authorisation implements Serializable {
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "case_type_id", nullable = false)
+    private CaseTypeLiteEntity caseType;
```
vlaurin: Do we need the link to the case type entity at all? I feel it would be more straightforward if we only had top-down links for those entities, so case type has an ACL, but the ACL doesn't have a link back to the case type.

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/EventLiteACLEntity.java
```diff
@@ -0,0 +1,21 @@
...
+import javax.persistence.*;
+import java.io.Serializable;
+
+@Table(name = "event_user_role")
+@Entity
+public class EventLiteACLEntity extends Authorisation implements Serializable {
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "event_id", nullable = false)
+    private EventLiteEntity event;
```
vlaurin: Same comment as for case type: Do we need the link to the event entity at all? I feel it would be more straightforward if we only had top-down links for those entities, so event has an ACL, but the ACL doesn't have a link back to the event.

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/CaseTypeLiteACLEntity.java
tzarsmango: removing reverse links to parent entities

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/EventLiteACLEntity.java
tzarsmango: removing reverse links to parent entities

PR #223, opened 2018-11-12
Rdm 2984 Same case type ID can be used across multiple jurisdictions

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-2984

### Change description ###
Added validation where I grab all case types from db and check if the case type reference being uploaded is not already there for some other jurisdiction to the one being uploaded and if it is then create case type non unique reference ValidationError.

```
[ ] Yes
[x] No
```

### DIFF
--- a/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
+++ b/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
@@ -12,6 +12,8 @@ public interface CaseTypeService {
 
     void createAll(JurisdictionEntity jurisdiction, Collection<CaseTypeEntity> caseTypes);
 
+    List<CaseType> findAll();
+
     List<CaseType> findByJurisdictionId(String jurisdictionId);
 
     Optional<CaseType> findByCaseTypeId(String id);
--- a/domain/src/test/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeServiceImplTest.java
+++ b/domain/src/test/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeServiceImplTest.java
@@ -1,8 +1,11 @@
 package uk.gov.hmcts.ccd.definition.store.domain.service.casetype;
 
+import org.assertj.core.util.Lists;
 import org.hamcrest.BaseMatcher;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
+import org.hamcrest.collection.IsEmptyCollection;
+import org.hamcrest.core.IsCollectionContaining;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Nested;
@@ -51,11 +54,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.same;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyZeroInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 import static org.mockito.hamcrest.MockitoHamcrest.argThat;
 import static uk.gov.hmcts.ccd.definition.store.repository.FieldTypeUtils.BASE_FIXED_LIST;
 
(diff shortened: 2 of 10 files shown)


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
```diff
@@ -12,6 +12,8 @@
 
     void createAll(JurisdictionEntity jurisdiction, Collection<CaseTypeEntity> caseTypes);
 
+    List<CaseType> findAll();
```
mario-paniccia: Are we retrieving all case types to do an exist? Isn't not performant and has high memory consumption? 
Can't we just ask the db whether a case type with a certain id already exists? (a count query on the db)

andrewfolga: We need all jurisdictions from the existing case types as we need to make sure we are not importing the same case type with different jurisdiction.

This is on import only so it should not affect performance that much.

andrewfolga: We are retrieving all most recent versions of case types here.

andrewfolga:
Answered.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
andrewfolga: Ok, to issue such count call I would have to make a call for each iteration of case types (see CaseTypeServiceImpl) hence it would be multiple calls (depending on how many case types there is defined in spreadsheet) instead of just one call at the beginning to get all.

mario-paniccia: that is ok, because a count query is very efficient. So even if you endup with 3 count queries one for each imported case type is always way more efficient than retrieving all the latest versions of all the case types

mario-paniccia: As a compromise, if we really want to take a shortcut, I think you can just use the following method we already have in the repository:
```    Optional<CaseTypeEntity> findCurrentVersionForReference(@Param("caseTypeReference") String caseTypeReference);
```
but this is still not as efficient as an exist method which returns a boolean, which is what I'm suggesting. I would avoid even the above method, for performance reasons

andrewfolga:
Changes applied. @mario-paniccia can you approve and I can start checkpointing?

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
```diff
@@ -12,6 +12,8 @@
 
     void createAll(JurisdictionEntity jurisdiction, Collection<CaseTypeEntity> caseTypes);
 
+    Optional<Boolean> findCaseTypeExists(String reference, String jurisdictionId);
```
mario-paniccia: why we need to return a Optional<Boolean> rather that a Boolean? This is confusing. Optional are returned in an api when the method might not return any result, as an alternative to returning null. But this is not our case. The case either exists or not. Can we change to Boolean?

mario-paniccia: can we call this method `caseTypeExists`? why are not doing a find here. A query is either an exist or a find.

domain/src/test/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeServiceImplTest.java
```diff
@@ -389,6 +398,34 @@ private EventEntity eventEntity(int order) {
...
+
+        private static final String CASE_TYPE_REFERENCE = "TestAddressBookCase";
+        private static final String JURISDICTION_REFERENCE = "TEST";
+
+        @Test
+        @DisplayName("Should return a present result if case type for jurisdiction other then given exist")
+        void shouldCallMapperAndReturnResult_whenRepositoryReturnsAnEntity() {
+            when(caseTypeRepository.caseTypeExists(CASE_TYPE_REFERENCE, JURISDICTION_REFERENCE)).thenReturn(2);
+
+            Optional<Boolean> caseTypeExists = classUnderTest.caseTypeExists(CASE_TYPE_REFERENCE, JURISDICTION_REFERENCE);
```
mario-paniccia: is this up to date? how come it's still returning an Optional?

domain/src/test/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeServiceImplTest.java
```diff
@@ -389,6 +398,34 @@ private EventEntity eventEntity(int order) {
...
+
+            assertThat(caseTypeExists.isPresent(), is(true));
+            verify(caseTypeRepository).caseTypeExists(CASE_TYPE_REFERENCE, JURISDICTION_REFERENCE);
+        }
+
+        @Test
+        @DisplayName("Should return an absent result if case type for jurisdiction other then given does no exist")
+        void shouldReturnEmptyList_whenRepositoryReturnsEmptyList() {
+            when(caseTypeRepository.caseTypeExists(CASE_TYPE_REFERENCE, JURISDICTION_REFERENCE)).thenReturn(0);
+            Optional<Boolean> caseTypeExists = classUnderTest.caseTypeExists(CASE_TYPE_REFERENCE, JURISDICTION_REFERENCE);
```
mario-paniccia: same here

domain/src/test/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeServiceImplTest.java
andrewfolga: Please refresh

domain/src/test/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeServiceImplTest.java
andrewfolga: Same as above

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
andrewfolga: done

andrewfolga: done

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
andrewfolga: done

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/casetype/CaseTypeService.java
```diff
@@ -12,6 +12,8 @@
 
     void createAll(JurisdictionEntity jurisdiction, Collection<CaseTypeEntity> caseTypes);
 
+    boolean caseTypeExists(String reference, String jurisdictionId);
```
mario-paniccia: we have to be careful here. If we look at this method, it looks like we are checking whether a case type exists with reference `reference` and jur `jurisdictionId`
But that is not the case. This method is doing the opposite. Is checking whether any case type with `reference` exists on any jurisdiction that is not `jurisdictionId`.

andrewfolga: Done - excludedJurisdictionReference used here

PR #314, opened 2019-02-26
RDM-3940 Updated event_case_field_complex_type label to text type

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-3940

### Change description ###
CaseEventToComplexTypes tab has a EventElementLabel we can use to override original label. This PR changes the label type in the DB to nullable Text.

```
[ ] Yes
[ X] No
```

### DIFF
--- a/repository/src/main/resources/db/changelog/db.changelog-rdm-3940.xml
+++ b/repository/src/main/resources/db/changelog/db.changelog-rdm-3940.xml
@@ -1,0 +1,13 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="rdm-3940" author="rafal.kalita@hmcts.net">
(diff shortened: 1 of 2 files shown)


repository/src/main/resources/db/changelog/db.changelog-rdm-3940.xml
```diff
@@ -0,0 +1,13 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="rdm-3940" author="rafal.kalita@hmcts.net">
+        <modifyDataType columnName="label"
+                        newDataType="text"
```
mario-paniccia: in other places where we define a label we are using a varchar with a max length specified. I assume that's to prevent very long labels that would cause issues on UI? Shall we do the same here?

rafalkalita: ok, will change to varchar(200)

rafalkalita: It is worth noticing that this label is an override, so I am dropping the NOT_NULL constraint. Tested locally and it will use the original label if no label provided.

PR #351, opened 2019-03-19
Read replicas

mario-paniccia:
RDM-4392

Useful links:
https://jdbc.postgresql.org/documentation/head/connect.html
https://github.com/jahe/spring-boot-multiple-datasources/tree/master/src/main/java/com/foobar/bar/repo
https://www.baeldung.com/spring-data-jpa-multiple-databases
https://www.ccampo.me/java/spring/2016/02/13/multi-datasource-spring-boot.html
https://github.com/jahe/spring-boot-multiple-datasources/tree/master/src/main/java/com/foobar/bar/repo

```
[ ] Yes
[x ] No
```


application/src/main/resources/application.properties
```diff
@@ -8,12 +8,19 @@ logging.level.org.springframework.web=INFO
...
 
-spring.datasource.url=jdbc:postgresql://${DEFINITION_STORE_DB_HOST:localhost}:${DEFINITION_STORE_DB_PORT:5000}/${DEFINITION_STORE_DB_NAME:ccd_definition}${DEFINITION_STORE_DB_OPTIONS:}
-spring.datasource.username=${DEFINITION_STORE_DB_USERNAME:ccd}
-spring.datasource.password=${DEFINITION_STORE_DB_PASSWORD:password}
-spring.datasource.type=com.zaxxer.hikari.HikariDataSource
-spring.datasource.hikari.maximum-pool-size=${DEFINITION_STORE_DB_MAX_POOL_SIZE:5}
-spring.datasource.hikari.connection-timeout=60000
+write.datasource.url=jdbc:postgresql://${DEFINITION_STORE_DB_HOST:localhost}:${DEFINITION_STORE_DB_PORT:5000}/${DEFINITION_STORE_DB_NAME:ccd_definition}\
+  {DEFINITION_STORE_DB_OPTIONS:}&targetServerType=master
+write.datasource.username=${DEFINITION_STORE_DB_USERNAME:ccd}
```
MSancaktutar: It could be better to follow a semantic hierarchy of properties, starting with app group and then followed by the app name and then etc etc . For example:
ccd.data-store-api.data-source.write-db.x
ccd.data-store-api.data-source.write-db.y.z

MSancaktutar: This comment may apply to similar other application parameters

mario-paniccia:
@gbenadikar what we want to do as part of this PR is not really improve performance of read only operations. 
We want to instead dispatch write operations only on the master db, and all read only operations to the master db and the slaves (the various read replicas we are going to create on Azure). Let me know how you think based on your suggestion I could simplify this PR. Cheers

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/VersionedDefinitionRepositoryDecorator.java
```diff
@@ -1,30 +1,38 @@
...
         s.setVersion(1 + version.orElse(0));
-        return repository.save(s);
+        return definitionWriteRepository.save(s);
     }
 
     @Override
     public <S extends T> List<S> saveAll(Iterable<S> iterable) {
+        List<S> result = new ArrayList<>();
         for (S s : iterable) {
             final Optional<Integer> version = repository.findLastVersion(s.getReference());
```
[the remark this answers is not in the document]
mario-paniccia: @gbenadikar good catch. I was aware of that problem, it's concerning me too. This approach would have problems if we do a write followed by a read which expects to see the write. I'm reviewing the code to see if we do this. The thing is, using this approach allows to introduce this feature without having to change much more code. I'll keep looking into this

PR #360, opened 2019-03-22
RDM-4402 Add endpoint to fetch all latest case type references.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4402

### Change description ###
Add endpoint to retrieve all latest case types.

```
[ ] Yes
[x] No
```


repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/QueryConstants.java
```diff
@@ -9,5 +9,8 @@
         "select c from CaseTypeEntity c where c.reference=:caseTypeReference " +
             "and c.version in (" + SELECT_MAX_CASE_TYPE_VERSION_NUMBER + ")";
 
+    public static final String SELECT_LATEST_CASE_TYPES_REFERENCES =
+        "select c.reference from CaseTypeEntity c where c.version in (select max(cm.version) from CaseTypeEntity as cm where cm.reference = c.reference)";
```
mario-paniccia: is it possible to reuse SELECT_MAX_CASE_TYPE_VERSION_NUMBER already defined query like the method above is doing?

rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/CaseDefinitionController.java
```diff
@@ -72,7 +72,17 @@ public CaseType dataCaseworkerIdAndJurisdictionIdCaseTypeGet(
         return caseRoleService.findByCaseTypeId(caseTypeId);
     }
 
-    @RequestMapping(value = "/data/jurisdictions/{jurisdiction_id}/case-type", method = RequestMethod.GET, produces = {"application/json"})
+    @RequestMapping(value = "/data/case-types-references", method = RequestMethod.GET, produces = {"application/json"})
```
mario-paniccia: Can we make this endpoint more REST? I've done a bit of googling and this one is a nice solution to me:
https://stackoverflow.com/questions/28681099/rest-getting-specific-property-of-resources-in-collection-only
`/data/case-types?field=reference`
What do you think?

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/QueryConstants.java
andrewfolga: I do not think so as my subquery is parameterless. It would be counter productive to add my subquery as a new one and use it as this would be rarely used I believe.

rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/CaseDefinitionController.java
andrewfolga: Sure we can

andrewfolga: Actually had another thought and wouldn't it suggest we can also call `/data/case-types` which should not be possible due to perf implications?

mario-paniccia: good point. We could add validation and return 400 bad request if they don't provide a field query param. Also we could return 400 for any field different than `reference`. i.e. for now we only support retrieving `reference`. 
Or, we could leave it as it is currently implemented. As you prefer

andrewfolga: Cool I will leave it than. Thanks 👍

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/QueryConstants.java
mario-paniccia: yes I see. 
By the way don't we need a `distinct` in `select c.reference from CaseTypeEntity`?

andrewfolga: Do not think so as the subquery works as a distinct already - it chooses one version for reference and the unique key in case_type table is - (reference, version).

PR #369, opened 2019-03-26
Rdm 4351 - Add new DynamicList basetype

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4351

Currently we have the 'FixedList' base type which services use to configure a list of options that the user can pick from. This data is fixed as per the configuration.

With the introduction of 'DynamicList' base type services will be able to update the list of options in the list dynamically via the AboutToStart and/or MidEvent call backs

This story is to create the 'DynamicList' base type and add to the definition store

```
[ ] Yes
[X] No
```


excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/parser/ListFieldTypeParser.java
```diff
@@ -81,6 +84,14 @@ public ListFieldTypeParser(ParseContext parseContext) {
         parseContext.addToAllTypes(fixedListType);
         result.addNew(fixedListType);
 
+        // Add as DynamicList with no list items as it will be dynamically added by the services
+        final FieldTypeEntity dynamicListType = new FieldTypeEntity();
+        dynamicListType.setBaseFieldType(fixedListBaseType);
```
mario-paniccia: shouldn't this be:
`dynamicListType.setBaseFieldType(dynamicListBaseType);`
?

excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/parser/ListFieldTypeParser.java
```diff
@@ -81,6 +84,14 @@ public ListFieldTypeParser(ParseContext parseContext) {
         parseContext.addToAllTypes(fixedListType);
         result.addNew(fixedListType);
 
+        // Add as DynamicList with no list items as it will be dynamically added by the services
+        final FieldTypeEntity dynamicListType = new FieldTypeEntity();
```
mario-paniccia: the ticket says:
```If a list has been been defined using 'FieldTypeParameter' when it is 'DynamicList' it configured list should be ignored
In this case do we really need this block of code here then?

File: application/src/test/java/uk/gov/hmcts/net/ccd/definition/store/excel/SpreadSheetImportTest.java [hunk 3]
```diff
@@ -282,11 +282,13 @@ private void assertFieldTypes() {
     }
 
     private void assertListFieldTypes(List<Map<String, Object>> fieldTypes) {
-
         assertThat(fieldTypes,
                    allOf(hasItem(allOf(hasColumn("jurisdiction_id", getIdForTestJurisdiction()),
                                        hasColumn("reference", "FixedList-marritalStatusEnum"),
                                        hasColumn("base_field_type_id", fieldTypesId.get("FixedList")))),
+                       hasItem(allOf(hasColumn("jurisdiction_id", getIdForTestJurisdiction()),
+                                       hasColumn("reference", "DynamicList-marritalStatusEnum"),
```
mario-paniccia: dynamic list most of the times will be configured with empty FieldTypeParameter. I think we should have that kind of test here. 
Here we have tested FieldTypeParameter=marritalStatusEnum which is ok, we can leave it. But we must have also a test for FieldTypeParameter empty.

repository/src/main/resources/db/changelog/db.changelog-rdm-4351.xml
```diff
@@ -0,0 +1,18 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
```
mario-paniccia: we need to add this new field type to the ElasticSearch mappings too. Look at:
application.yml
After you do that, can you make sure you are able to import a spreadsheet with ElasticSearch turned on? If you try that at the moment, you will see
the import fails because on an ElasticSearch error. We need to fix that

application/src/test/java/uk/gov/hmcts/net/ccd/definition/store/excel/SpreadSheetImportTest.java
```diff
@@ -282,11 +282,13 @@ private void assertFieldTypes() {
...
+                                       hasColumn("base_field_type_id", fieldTypesId.get("FixedList")))),
                          hasItem(allOf(hasColumn("jurisdiction_id", getIdForTestJurisdiction()),
                                        hasColumn("reference", "MultiSelectList-marritalStatusEnum"),
```
mario-paniccia: furthermore, is it possible to add some more unit tests rather than relying mostly on this test here which is an end to end test

excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/parser/ListFieldTypeParser.java
mohammedanas: correct thanks for spotting it

repository/src/main/resources/db/changelog/db.changelog-rdm-4351.xml
mohammedanas: Added them

excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/parser/ListFieldTypeParser.java
mohammedanas: Removed check for fieldtypeparameter changes for DynamicList

application/src/test/java/uk/gov/hmcts/net/ccd/definition/store/excel/SpreadSheetImportTest.java
mohammedanas: Yes tested with both cases and added to IT excel sheet

excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/parser/ListFieldTypeParser.java
```diff
@@ -28,18 +29,20 @@
...
     private final FieldTypeEntity multiListBaseType;
+    private final FieldTypeEntity dynamicListBaseType;
 
     public ListFieldTypeParser(ParseContext parseContext) {
         this.parseContext = parseContext;
         fixedListBaseType = parseContext.getBaseType(FIXED_LIST_TYPE).orElseThrow(() -> new InvalidImportException(NO_BASE_TYPE_FOUND + FIXED_LIST_TYPE));
         fixedRadioListBaseType = parseContext.getBaseType(FIXED_RADIO_LIST_TYPE).orElseThrow(() -> new InvalidImportException(NO_BASE_TYPE_FOUND + FIXED_RADIO_LIST_TYPE));
         multiListBaseType = parseContext.getBaseType(MULTI_LIST_TYPE).orElseThrow(() -> new InvalidImportException(NO_BASE_TYPE_FOUND + MULTI_LIST_TYPE));
-
+        dynamicListBaseType = parseContext.getBaseType(DYNAMIC_LIST_TYPE).orElseThrow(() -> new InvalidImportException(NO_BASE_TYPE_FOUND + DYNAMIC_LIST_TYPE));
```
mario-paniccia: the ticket says: 
```If a list has been been defined using 'FieldTypeParameter' when it is 'DynamicList' it configured list should be ignored```
given this requirement, can we just remove this line here? Because we have to ignore this info anyway

mario-paniccia: perhaps removing that line will fix RDM-4634?

PR #372, opened 2019-04-01
RDM-4392: Read replicas db

mario-paniccia:
https://tools.hmcts.net/jira/browse/RDM-4392
https://jdbc.postgresql.org/documentation/head/connect.html
https://github.com/ttddyy/datasource-proxy

Configure Def Store with 2 data sources one pointing to the Master DB the other one to load balance between replicas.
If read replica Data source is enabled, all db requests are by default directed to the read replicas (read replicas made the default because Def Store has only
1 endpoint that produces write operations, the import endpoint. All other endpoints are read endpoints). 
To route write/read flows to the master DB, use the @RouteToMaster annotation

```
[ ] Yes
[ x] No
```

### DIFF
--- a/application/build.gradle
+++ b/application/build.gradle
@@ -22,6 +22,7 @@ dependencies {
     compile "org.springframework.boot:spring-boot-starter-actuator"
     compile "io.springfox:springfox-swagger2:${springfoxSwaggerVersion}"
     compile "io.springfox:springfox-swagger-ui:${springfoxSwaggerVersion}"
+    compile group: 'net.ttddyy', name: 'datasource-proxy', version: '1.5.1'
     runtime group: 'org.postgresql', name: 'postgresql', version: '42.2.5'
     runtime "org.liquibase:liquibase-core"
     testCompile project(":app-insights").sourceSets.main.output
--- a/application/src/main/resources/application.properties
+++ b/application/src/main/resources/application.properties
@@ -8,12 +8,13 @@ logging.level.org.springframework.web=INFO
 logging.level.uk.gov.hmcts.ccd=INFO
 logging.level.org.elasticsearch.client=INFO
 
-master.datasource.url=jdbc:postgresql://${DEFINITION_STORE_DB_HOST:localhost}:${DEFINITION_STORE_DB_PORT:5000}/${DEFINITION_STORE_DB_NAME:ccd_definition}$\
+spring.datasource.url=jdbc:postgresql://${DEFINITION_STORE_DB_HOST:localhost}:${DEFINITION_STORE_DB_PORT:5000}/${DEFINITION_STORE_DB_NAME:ccd_definition}$\
   {DEFINITION_STORE_DB_OPTIONS:}
-master.datasource.username=${DEFINITION_STORE_DB_USERNAME:ccd}
-master.datasource.password=${DEFINITION_STORE_DB_PASSWORD:password}
-master.datasource.hikari.maximum-pool-size=${DEFINITION_STORE_DB_MAX_POOL_SIZE:5}
-master.datasource.hikari.connection-timeout=60000
+spring.datasource.username=${DEFINITION_STORE_DB_USERNAME:ccd}
+spring.datasource.password=${DEFINITION_STORE_DB_PASSWORD:password}
+spring.datasource.type=com.zaxxer.hikari.HikariDataSource
+spring.datasource.hikari.maximum-pool-size=${DEFINITION_STORE_DB_MAX_POOL_SIZE:5}
+spring.datasource.hikari.connection-timeout=60000
 
 replicas.datasource.url=jdbc:postgresql://${DEFINITION_STORE_DB_HOST:localhost}:${DEFINITION_STORE_DB_PORT:5000}/${DEFINITION_STORE_DB_NAME:ccd_definition}$\
   {DEFINITION_STORE_DB_OPTIONS:}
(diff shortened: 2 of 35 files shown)


application/build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
     compile "org.springframework.boot:spring-boot-starter-actuator"
     compile "io.springfox:springfox-swagger2:${springfoxSwaggerVersion}"
     compile "io.springfox:springfox-swagger-ui:${springfoxSwaggerVersion}"
+    compile group: 'net.ttddyy', name: 'datasource-proxy', version: '1.5.1'
```
mario-paniccia: helps debugging the sql calls

repository/src/test/java/uk/gov/hmcts/ccd/definition/store/database/RoutingDataSourceTest.java
```diff
@@ -0,0 +1,30 @@
+package uk.gov.hmcts.ccd.definition.store.database;
```
mario-paniccia: I did add also an integration test for the /import controller to make sure all operations are sent to the master db. Unfortunately any integration test won't work. Because tests are single threaded. And this change relies on ThreadLocal. Which would require the test to use separate threads to work. So, in short, no idea how to test this with an integration test.

application/src/main/resources/application.properties
```diff
@@ -15,6 +15,16 @@ spring.datasource.type=com.zaxxer.hikari.HikariDataSource
 spring.datasource.hikari.maximum-pool-size=${DEFINITION_STORE_DB_MAX_POOL_SIZE:5}
 spring.datasource.hikari.connection-timeout=60000
 
+replicas.datasource.enabled=${REPLICAS_DATASOURCE_ENABLED:false}
+replicas.datasource.url=jdbc:postgresql://${DEFINITION_STORE_REPLICA_HOSTS:localhost}:${DEFINITION_STORE_DB_PORT:5000}/\
+  ${DEFINITION_STORE_DB_NAME:ccd_definition}${DEFINITION_STORE_DB_OPTIONS:?}&targetServerType=any&loadBalanceHosts=true
```
mario-paniccia: we send read request to all databases, master and the replicas

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/database/RoutingDataSource.java
```diff
@@ -0,0 +1,26 @@
...
+        ctx.remove();
+        ctx.set(Route.REPLICA);
+    }
+
+    public static void setMasterRoute() {
+        ctx.set(Route.MASTER);
+    }
+
+    @Override
+    protected Object determineCurrentLookupKey() {
```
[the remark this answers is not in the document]
mario-paniccia: done

rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/UserRoleController.java
```diff
@@ -39,6 +40,7 @@
         @ApiResponse(code = 205, message = "User role is updated successfully and the user agent SHOULD reset the document view"),
         @ApiResponse(code = 409, message = "Bad request, for example, incorrect data")
     })
+    @RouteToMasterDB
```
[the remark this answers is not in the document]
mario-paniccia: on Repository layer things won't work. We could end up with inconsistencies. 
Imagine a controller uses 2 repositories Repo A and Repo B. Does a write with Repo A and a read with Repo B, and expects to see the write. 
We can annotate RepoB write method with @RouteToMasterDB. But we cannot annotate Repo B read method with @RouteToMasterDB. Because in general we want reads to go also to replicas. So we cannot reuse this method in other scenarios if we annotated it.
With this setup, the controller won't be able to see the write. And there will be inconsistencies.
Same issue for Service layers. A controller could use multiple Services

repository/src/test/java/uk/gov/hmcts/ccd/definition/store/database/RoutingDataSourceTest.java
mario-paniccia: is there an easy way to do that Mutlu without writing complex multhithread handling code in a test? If I spawn a new thread in a test, then the main thread has to wait for the spawned thread to complete before marking the test as completed. Things gets messy.
I hoped JUNIT offered support for this but couldn't find anything.

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/database/MasterDBConnectionInterceptor.java
```diff
@@ -0,0 +1,29 @@
...
+        //empty comment to please Sonar
+    }
+
+    @Around("@annotation(routeToMasterDB)")
+    public Object proceed(ProceedingJoinPoint pjp, RouteToMasterDB routeToMasterDB) throws Throwable {
+        try {
+            RoutingDataSource.setMasterRoute();
+            return pjp.proceed();
+        } finally {
+            RoutingDataSource.clearMasterRoute();
```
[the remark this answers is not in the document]
mario-paniccia: issue with that is that is not easy to do. I would have to introduce another state variable to keep track on the previous state. Should that variable be ThreadLocal as well? I think so. Things gets complicated. 
For the usage we are doing of the annotation, I don't think it's worth it right now. 
With the current usage, where we just place the @RouteToMaster db annotation on the outmost controller methods only, this change is not required. 
I would prefer simplicity for now

PR #379, opened 2019-04-08
RDM-4426 Complex Element *only* on list screen - Import Validation

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-4426

### Change description ###
Validation for ListElementCode in:
- SearchInputFields
- SearchResultFields
- WorkBasketInputFields
- WorkBasketResultFields

```
[ ] Yes
[X ] No
```


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
```diff
@@ -59,4 +54,36 @@ public void createDisplayGroups(List<DisplayGroupEntity> displayGroups) {
...
+    private void validate(List<GenericLayoutEntity> genericLayouts) {
+        ValidationResult result = new ValidationResult();
+        for (GenericLayoutEntity genericLayoutEntity : genericLayouts) {
+            for (GenericLayoutValidator validator : genericLayoutValidators) {
+                result.merge(validator.validate(genericLayoutEntity));
+            }
+        }
+
+        for (GenericLayoutEntity genericLayoutEntity : genericLayouts) {
+            genericLayouts.stream()
```
rafalkalita: Validating in code instead of unique constraint in the db.

repository/src/main/resources/db/changelog/db.changelog-rdm-4426.xml
```diff
@@ -0,0 +1,32 @@
...
+            <column name="case_field_element_path" type="varchar(300)"/>
+        </addColumn>
+        <addColumn tableName="workbasket_case_field">
+            <column name="case_field_element_path" type="varchar(300)"/>
+        </addColumn>
+        <addColumn tableName="workbasket_input_case_field">
+            <column name="case_field_element_path" type="varchar(300)"/>
+        </addColumn>
+
+        <dropUniqueConstraint constraintName="unique_search_input_case_field_case_type_id_case_field_id"
```
rafalkalita: Dropping the constraints and moved validation to the code, so it will be in one place. The alternative was to create the unique constraints with three elements, of which one can be null.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
```diff
@@ -59,4 +54,36 @@ public void createDisplayGroups(List<DisplayGroupEntity> displayGroups) {
...
+
+    private void validate(List<GenericLayoutEntity> genericLayouts) {
+        ValidationResult result = new ValidationResult();
+        for (GenericLayoutEntity genericLayoutEntity : genericLayouts) {
+            for (GenericLayoutValidator validator : genericLayoutValidators) {
+                result.merge(validator.validate(genericLayoutEntity));
+            }
+        }
+
+        for (GenericLayoutEntity genericLayoutEntity : genericLayouts) {
```
mario-paniccia: can this validation for uniqueness be moved into a new GenericLayoutValidator?

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
```diff
@@ -59,4 +54,36 @@ public void createDisplayGroups(List<DisplayGroupEntity> displayGroups) {
...
+            genericLayouts.stream()
+                .filter(e -> e != genericLayoutEntity)
```
mario-paniccia: this code in terms of performance is O(n2). Any way we can improve the performance?

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
```diff
@@ -59,4 +54,36 @@ public void createDisplayGroups(List<DisplayGroupEntity> displayGroups) {
...
+                .filter(e -> e.getCaseField().getReference().equals(genericLayoutEntity.getCaseField().getReference()))
+                .filter(e -> e.getCaseType().getReference().equals(genericLayoutEntity.getCaseType().getReference()))
```
mario-paniccia: can we move this chain of filters into method with a self explanatory name?

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/genericlayout/GenericLayoutEntityValidatorImpl.java
```diff
@@ -34,6 +39,34 @@ public ValidationResult validate(GenericLayoutEntity entity) {
                         (entity.getCaseType() != null ? entity.getCaseType().getReference() : "")
                     ), entity)
             );
+        } else {
+            if (entity.getCaseFieldElementPath() != null) {
```
mario-paniccia: too long block of code. Can we break it down to multiple private methods please

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/genericlayout/GenericLayoutEntityValidatorImpl.java
```diff
@@ -34,6 +39,34 @@ public ValidationResult validate(GenericLayoutEntity entity) {
...
+                if (entity.getCaseField().isComplexFieldType() || entity.getCaseField().isCollectionFieldType()) {
+                    List<CaseFieldEntity> caseFields = entity.getCaseType().getCaseFields();
+
+                    List<String> allPaths = CaseFieldEntityUtil.buildDottedComplexFieldPossibilities(caseFields);
```
mario-paniccia: can we avoid hardcoding CaseFieldEntityUtil and use dependency injection instead?

mario-paniccia: is there a potential performance issue here? We are building all the paths multiple times for a same case type

rafalkalita: No performance issue here. It takes from 1 to 2 ms in average for one execution if ListElementCode is present. Additionally our definition import happens rarely.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
rafalkalita: Moved all validation to GenericLayoutEntityValidatorImpl

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
rafalkalita: There is a handful of GenericLayoutEntities and executed on definition import. I think this is not an issue.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
```diff
@@ -30,12 +30,17 @@
     private DisplayGroupRepository displayGroupRepository;
```
mario-paniccia: DisplayGroupAdapterService is growing too much. Long methods. Can we do some heavy refactorings please? Introduce multiple self explanatory private methods for example

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
```diff
@@ -59,7 +64,7 @@ public WizardPageCollection findWizardPagesByCaseTypeId(String caseTypeReference
 
                     List<String> allSubTypePossibilities =
```
mario-paniccia: I might be wrong but `allSubTypePossibilities` is initialised on line 65 but then used only on line 102. Can we move it close to the point of usage please

rafalkalita: It is initialized outside of the forEach loop to avoid unnecessary cycles.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
rafalkalita: done

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
```diff
@@ -45,74 +53,100 @@ public WizardPageCollection findWizardPagesByCaseTypeId(String caseTypeReference
...
+        wizardPage.setId(displayGroupEntity.getReference());
+        wizardPage.setLabel(displayGroupEntity.getLabel());
+        wizardPage.setOrder(displayGroupEntity.getOrder());
+
+        Optional.ofNullable(displayGroupEntity.getWebhookMidEvent()).ifPresent(webhookEntity -> {
+            wizardPage.setCallBackURLMidEvent(webhookEntity.getUrl());
+            wizardPage.setRetriesTimeoutMidEvent(webhookEntity.getTimeouts());
+        });
+
+        List<String> allSubTypePossibilities = createAllSubtypeLeafCombinations(displayGroupEntity);
```
mario-paniccia: any reason to create the variable here and pass it to `createWizardCaseField` rather than creating it inside `createWizardCaseField` itself?

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/displaygroup/TabShowConditionValidatorImpl.java
```diff
@@ -43,7 +46,7 @@ public ValidationResult validate(DisplayGroupEntity thisDisplayGroup, List<Displ
                 return validationResult;
             }
 
-            List<String> allSubTypePossibilities = CaseFieldEntityUtil
+            List<String> allSubTypePossibilities = caseFieldEntityUtil
```
mario-paniccia: this block of code is very similar if not identical to the one below and the one in PageShowConditionValidatorImpl. We are introducing some duplication. I think we should extract this into its own class, called something like FieldsPathValidator or sim

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/eventcasefieldcomplextype/EventComplexTypeShowConditionValidatorImpl.java
```diff
@@ -50,7 +53,7 @@ public ValidationResult validate(EventComplexTypeEntity eventCaseFieldEntity,
         List<EventCaseFieldEntity> allEventCaseFieldEntitiesForEventCase =
             eventCaseFieldEntityValidationContext.getAllEventCaseFieldEntitiesForEventCase();
 
-        List<String> allSubTypePossibilities = CaseFieldEntityUtil.buildDottedComplexFieldPossibilities(
```
mario-paniccia: and this block of code is very similar to the one in EventCaseFieldShowConditionValidatorImpl. I think `FieldsPathValidator` class is a must now. Too much duplication

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/genericlayout/GenericLayoutEntityValidatorImpl.java
```diff
@@ -1,68 +1,159 @@
...
 import uk.gov.hmcts.ccd.definition.store.repository.entity.GenericLayoutEntity;
 
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
 @Component
 public class GenericLayoutEntityValidatorImpl implements GenericLayoutValidator {
```
mario-paniccia: I would split this into multiple validators. What's the point of creating an interface  `GenericLayoutValidator` if then we have all the different validations happening in 1 subclass only?

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
rafalkalita: Because it is out of displayGroupEntity.getDisplayGroupCaseFields().stream() so it gets computed once.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/genericlayout/GenericLayoutEntityValidatorImpl.java
rafalkalita: Done

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/LayoutServiceImpl.java
```diff
@@ -59,4 +52,17 @@ public void createDisplayGroups(List<DisplayGroupEntity> displayGroups) {
         }
         displayGroupRepository.saveAll(displayGroups);
     }
+
+    private void validate(List<GenericLayoutEntity> genericLayouts) {
+        ValidationResult result = new ValidationResult();
+
+        for (GenericLayoutValidator validator : genericLayoutValidators) {
```
mario-paniccia: any reason why we changed the way we did validation? i.e. before we were looping through the fields first and applying for each field the validators. And this is what we do for all the validations also on other types of fields. But here we swapped it. We loop through the validators, and then each validator loops through all the fields. I think the other way made more sense

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/validation/genericlayout/GenericLayoutEntityOrderValidatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+public class GenericLayoutEntityOrderValidatorImpl implements GenericLayoutValidator {
+
+    private static final String ERROR_MESSAGE_INVALID_NUMBER_WITH_CASE_FIELD =
+        "DisplayOrder '%d' needs to be a valid integer for row with label '%s', case field '%s'";
+
+    @Override
+    public ValidationResult validate(List<GenericLayoutEntity> genericLayoutEntities) {
+        final ValidationResult validationResult = new ValidationResult();
+
+        for (GenericLayoutEntity entity : genericLayoutEntities) {
```
mario-paniccia: this looping logic should have been written once and for all validators in LayoutServiceImpl.java , rather than duplicated in each validator. And a validator should only validate 1 single field.

tzarsmango: Agree to this one. We write entity validators and each validates validate an entity.We may need to send all entities to the validator method for validating duplicity etc..

tzarsmango:
This PR will have some code conflicts with [this PR](https://github.com/hmcts/ccd-definition-store-api/pull/386). So watch out for changes if it's merged before.

PR #392, opened 2019-04-29
RDM-4272-1: Update the labels for Address Line 2 and 3

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-4272

### Change description ###
Update the labels for Address Line 2 and 3

```
[ ] Yes
[x] No
```


repository/src/main/resources/db/changelog/db.changelog-rdm-4272.xml
```diff
@@ -0,0 +1,19 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+
+    <changeSet id="rdm-4272" author="karamath.mohamed@hmcts.net">
+
+        <!-- Update the label for Address line 2 and 3 in the complex_field table -->
+        <update tableName="complex_field">
+            <column name="label" value="Address Line 2" />
+            <where>reference = 'AddressLine2'</where>
```
gbenadikar: should there be a check in the where clause either for specific complex field type or label null/blank, before setting the label for all fields with reference='AddressLine2'? The same reference may have been used for non-base type complex address fields. We don't want them overwritten with this statement.

MKaramath: Updated the where clause to add the label blank check as discussed

PR #401, opened 2019-05-15
Fix spreadsheet parser: string value of the cell

doncem:
### Change description ###

Reference: hmcts/bulk-scan-shared-infrastructure#108

Numerously tested while doing reference above and simply using `toString` which is fully implemented by the library is enough (BTW, can even get rid of `BLANK` check before. Only now thought of it). The main problem is complex string retrieval - it is not always just `_cell.getV()`.

Bottomline any import caused `NullPointerException` without any knowledge why as last cause was not exposed in server stacktrace.

Bonus overview/summary: there was a moment I missed one tiny configuration and import just looked like stuck at one place. Only figured out my mistake once provided logging for `ValidationException` in the `ControllerAdvice`

One more bonus: including tiny merge in try-with-resources so `SpreadsheetParser` doesn't have suppose "not used variable" warning. `Workbook` itself is `Closeable` so can squeeze that in and be safe everything is closed properly

```
[ ] Yes
[x] No
```


doncem:
Forks not working with jenkins pipeline. Will create hmcts copy in a bit

PR #408, opened 2019-05-30
Log import validation errors

mario-paniccia:
On the new ITHC env we noticed that for some reason sometimes the import validation errors are not returned to the UI and are not logged. This makes it difficult to understand what the issue is because the import just fails silently. 
This quick PR adds logging of validation errors 

```
2019-05-29T14:24:13.243 INFO  [http-nio-1732-exec-9] u.g.h.c.d.s.d.service.casetype.CaseTypeServiceImpl validationError: Invalid UserRole for case type 'AAT_PRIVATE', case state 'TODO'
2019-05-29T14:24:13.243 INFO  [http-nio-1732-exec-9] u.g.h.c.d.s.d.service.casetype.CaseTypeServiceImpl validationError: Invalid UserRole for case type 'AAT_PRIVATE', case state 'TODO'
```

```
[ ] Yes
[ x] No
```


dwaynebailey:
LGTM - how easy would it be to actually report the role that is missing? At the moment you're left guessing which one is missing.  Looks easy enough using `entity.getRoleName()` in the Exception.  Not a blocker.

PR #428, opened 2019-07-01
RDM-5242 Overrides fix when the fields in one event have similar prefix

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-5242

### Change description ###
Adding a dot  character when filtering override possibilities to avoid a bug when names have similar beginning.

```
[ ] Yes
[ X] No
```

### DIFF
--- a/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
+++ b/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
@@ -123,7 +123,7 @@ public class DisplayGroupAdapterService {
             .collect(Collectors.toList());
 
         return allSubTypePossibilities.stream()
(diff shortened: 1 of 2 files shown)


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/display/DisplayGroupAdapterService.java
```diff
@@ -123,7 +123,7 @@ private WizardPageField createWizardCaseField(final DisplayGroupEntity displayGr
             .collect(Collectors.toList());
 
         return allSubTypePossibilities.stream()
-            .filter(e -> e.startsWith(displayGroupCaseFieldEntity.getCaseField().getReference()))
+            .filter(e -> e.startsWith(displayGroupCaseFieldEntity.getCaseField().getReference() + '.'))
```
rafalkalita: This was failing when in one event we have got claimant and claimantWorkAddress fields in Ethos definition. The overrides from claimantWorkAddress were added to claimant overrides causing front-end to show error message `Cannot find leaf for caseFieldId`

PR #463, opened 2019-10-31
Performance improvements for CaseType entity loading

banderous:
https://tools.hmcts.net/jira/browse/RDM-6246

### Change description ###
Performance improvements for the loading of CaseType entities

This was developed against a test data set created by importing Divorce's case definition 2000 times, with load time improving from ~7 seconds to ~250ms with these changes. Loading of /display/tab-structure has also improved from ~9 seconds to ~ 400ms for the same dataset, although not the original objective for this PR. Other operations may also see improvements.

The biggest win has been migrating the three webhooks on event (start, pre and post submit) to the new event_webhook table. Having these webhooks as columns resulted in a very expensive fetch with webhook being joined to event 3 times, but the biggest impact was the non-performant SQL generated by Hibernate when joining through the event table.

Migrating webhook timeouts to an integer[] column on webhook provided another significant boost; modelling these relationally was expensive given that for a typical webhook these might only take a dozen bytes of storage and we don't need the flexibility of a relation.

In addition to our automated integration tests I have tested the correctness of the changes by comparing the JSON request response before and after these changes to ensure it is semantically identical (using jq as described [here](https://stackoverflow.com/questions/31930041/using-jq-or-alternative-command-line-tools-to-compare-json-files)

Note that I have deleted some unit tests that were broken by these changes - I don't believe these unit tests were testing anything beyond the type checking the java compiler already provides. I had to update an integration test when making these changes, so we have coverage of saving and loading with the schema changes.

```
[ ] Yes
[x] No
```


repository/src/main/resources/db/changelog/db.changelog-rdm-6246.2.xml
```diff
@@ -0,0 +1,48 @@
...
+the previous code (webhook_timeouts should be dropped on a follow up change).
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="rdm-6246.add-timeouts-column" author="alex.mcausland@hmcts.net">
+        <sql>
+            alter table webhook
+            add column timeouts integer[]
+            /* Disallow null values in the array*/
```
mario-paniccia: I think this should be done at the application level not the db like we do other validations. Check the *Validator classes

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.3.xml
```diff
@@ -0,0 +1,54 @@
...
+                where webhook_pre_submit_id is not null
+                union select id, webhook_post_submit_id, 'POST_SUBMIT'
+                from event
+                where webhook_post_submit_id is not null
+            );
+            <!--
+            Prompt Postgres to update its statistics for the table since we have inserted a lot of rows.
+            Otherwise Postgres may generate inefficient query plans until the next time its autovacuum daemon runs.
+            -->
+            analyze event_webhook;
```
mario-paniccia: wonder whether it'll be safer to do all the db preparation and data migration outside of Def Store deployment in a preliminary stage. We could provide the migration scripts to our DBA Alliu and ask him to do the migration.That would give us more control? I'm a bit concerned on what might happen to the prod Db if there's some issue during the Def Store deployment while data migration is in progress. Maybe I'm being too cautious but let's have a chat about this.

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/EventEntity.java
```diff
@@ -294,4 +288,15 @@ public Boolean getCanSaveDraft() {
     public void setCanSaveDraft(Boolean canSaveDraft) {
         this.canSaveDraft = canSaveDraft;
     }
+
+    private void setWebhook(EventStage state, WebhookEntity webhook) {
```
mario-paniccia: `stage` rather than `state`?

mario-paniccia:
@banderous is such a big perf improvement also happening on definitions that don't have any callback configured?

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.2.xml
```diff
@@ -0,0 +1,48 @@
...
+            check (array_position(timeouts, NULL) IS NULL);
```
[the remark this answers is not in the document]
banderous: I think we covered this one on slack; this is equivalent to the previous db constraint making timeouts non null - just a little harder to read :)

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.2.xml
```diff
@@ -0,0 +1,48 @@
...
+        set timeouts = times.timeouts
+        from (
+          select
+            webhook_id,
+            array_agg(timeout order by index) as timeouts
+          from webhook_timeout
+          group by webhook_id
+        ) times
+        where webhook.id = times.webhook_id;
+        </sql>
```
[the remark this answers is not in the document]
banderous: Yes good idea.

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.3.xml
```diff
@@ -0,0 +1,54 @@
...
+There were previously three columns on event containing that event's webhooks.
+This takes a relational approach to enable better fetching strategies.
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet id="rdm-6246.create_event_stage_enum" author="alex.mcausland@hmcts.net">
+        <!--This Enum models the different stages at which an event can fire a webhook-->
+        <sql>
+            create type event_stage as enum('START', 'PRE_SUBMIT', 'POST_SUBMIT');
```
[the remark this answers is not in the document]
banderous: Yes I was wondering about the appropriate terminology - I'll do the rename.

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.3.xml
```diff
@@ -0,0 +1,54 @@
...
+        <rollback>drop table event_webhook;</rollback>
+    </changeSet>
+    <changeSet id="rdm-6246.populate_event_webhook_table" author="alex.mcausland@hmcts.net">
+        <!--Populate the newly created table.-->
+        <!--For every event we need to insert a row for every start,-->
+        <!--presubmit and postsubmit webhook to link the event to its webhook-->
+        <sql>
+            insert into event_webhook(event_id, webhook_id, event_stage)
+            (
+                select id as event_id, webhook_start_id as webhook_id, 'START'::event_stage
```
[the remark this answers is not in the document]
banderous: The first select defines the column types of the query - I need to quote the enum type here or postgres will complain about a string to enum conversion. The subsequent selects are automatically converted to the enum type.

I could suffix all the enums which might make it more readable

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.3.xml
banderous: These migrations should be fail safe since they are nondestructive; the existing schema should be preserved so the current App code can continue to run during and after the migration (and can continue to run unaffected if any part of the migration fails).

For sure we cannot be too cautious, we should go into it in depth next week.

banderous:
@mario-paniccia I have tested across different case types using a Demo definition store db dump, where I'm seeing a 95 - 99% reduction in query times for both getting a case type by ID and fetching case display tabs.

I also repeated my functional equivalence tests and saw identical responses across Probate, Divorce, CMC and Financial remedy.

Some queries are still taking 300ms and there may be scope for further reductions; the FieldType entity has two self referencing columns which appear to be inefficient for Hibernate to fetch.

Also, of the 340k field_type rows in Demo, only 14 rows actually have children, which makes me think there may be a different/faster way to model this.

  | (develop) GetCaseById (ms) | (develop) DisplayTabs (ms) | (RDM-6246) GetCaseById (ms) | (RDM-6246) DisplayTabs (ms) | GetCase % reduction | DisplayTabs % reduction
-- | -- | -- | -- | -- | -- | --
DIVORCE | 4684 | 4707 | 187 | 161 | 96.01% | 96.58%
GrantOfRepresentation | 5150 | 5140 | 200 | 225 | 96.12% | 95.62%
FinancialRemedyMVP2 | 4493 | 4455 | 134 | 139 | 97.02% | 96.88%
MoneyClaimCase | 3728 | 4303 | 165 | 163 | 95.57% | 96.21%
Benefit | 9646 | 9144 | 333 | 349 | 96.55% | 96.18%
CARE_SUPERVISION_EPO | 7723 | 7814 | 160 | 136 | 97.93% | 98.26%
TRIB_MVP_3_TYPE | 6508 | 6530 | 109 | 108 | 98.33% | 98.35%
SEN_MVP_TYPE | 4181 | 4155 | 105 | 110 | 97.49% | 97.35%
Divorce_ExceptionRecord | 2784 | 2814 | 78 | 71 | 97.20% | 97.48%
MoneyClaimCaseTest | 4363 | 4458 | 166 | 171 | 96.20% | 96.16%
Shared_Storage_DRAFTType | 7441 | 7462 | 73 | 84 | 99.02% | 98.87%
Caveat | 5465 | 4730 | 90 | 118 | 98.35% | 97.51%
Demo | 9197 | 8779 | 304 | 330 | 96.69% | 96.24%
TRIB_MVP_1_TYPE | 6646 | 6548 | 122 | 112 | 98.16% | 98.29%
WillLodgement | 4205 | 4166 | 72 | 74 | 98.29% | 98.22%
StandingSearch | 4163 | 4257 | 70 | 71 | 98.32% | 98.33%

repository/src/main/resources/db/changelog/db.changelog-rdm-6246.3.xml
mario-paniccia: what we could do is asking Alliu, our DBA, to do a full db backup just before the deployment. That should be enough I think. I added a comment on the ticket in this regard.

mario-paniccia:
These are remarkable results @banderous, great job there! 🥇

PR #467, opened 2019-11-04
Address current CVEs and clean up suppressions

banderous:
https://tools.hmcts.net/jira/browse/RDM-6331

### Change description ###
Address current CVEs by bumping dependency versions and clean up redundant CVE suppressions.

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,7 +124,7 @@ allprojects {
     apply plugin: 'java'
 
     ext {
-        hibernateVersion = '5.2.16.Final'
+        hibernateVersion = '5.2.18.Final'
         elasticSearchVersion = '6.8.4'
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -91,6 +91,8 @@ dependencyManagement {
     }
 
     dependencies {
+        // Versions prior to 24.1.1 vulnerable to CVE-2018-10237
+        dependency 'com.google.guava:guava:24.1.1-jre'
         // CVE-2019-0232, CVE-2019-0199  - Java and Command Line injections in Windows
         dependencySet(group: 'org.apache.tomcat.embed', version: '9.0.21') {
             entry 'tomcat-embed-core'
--- a/build.gradle
+++ b/build.gradle
@@ -124,7 +124,7 @@ allprojects {
 
     ext {
         hibernateVersion = '5.2.16.Final'
-        elasticSearchVersion = '6.4.2'
+        elasticSearchVersion = '6.8.4'
     }
 
     dependencies {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -120,14 +124,14 @@ allprojects {
...
-        elasticSearchVersion = '6.4.2'
+        hibernateVersion = '5.2.18.Final'
+        elasticSearchVersion = '6.8.4'
     }
 
     dependencies {
         testCompile "org.springframework.boot:spring-boot-starter-test:${springBootVersion}"
 
-        compile "com.fasterxml.jackson.core:jackson-databind:2.9.10"
+        compile "com.fasterxml.jackson.core:jackson-databind:2.9.10.1"
```
smathangi: may be bump to latest 2.10.0?

banderous: Since this is a transitive dependency I wanted to keep the difference in version numbers as small as possible to minimise the possibility of an incompatibility.

We don't directly depend on jackson databind so I have replaced this direct dependency with a line in the spring dependency configuration section to ensure the correct version is used, if you're happy with this approach?

build.gradle
```diff
@@ -120,14 +127,13 @@ allprojects {
     apply plugin: 'java'
 
     ext {
-        hibernateVersion = '5.2.16.Final'
-        elasticSearchVersion = '6.4.2'
+        hibernateVersion = '5.2.18.Final'
+        elasticSearchVersion = '6.8.4'
```
mario-paniccia: Alex do we need to bump up this? The reason why I'm asking is that the ES backend we use is on version 6.4.2 so ideally it would be better to use a client of corresponding version

banderous: There are open CVEs for 6.4.2; https://nvd.nist.gov/vuln/detail/CVE-2018-17244 and https://nvd.nist.gov/vuln/detail/CVE-2019-7611

If we can demonstrate that we are not affected we could continue to suppress, but long term we will need to upgrade - can we upgrade the backend too?

banderous: According to the [elasticsearch guide](https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/client.html) the client should work with any server on the same major version, though as you say ideally client and server versions should match:

> The client must have the same major version (e.g. 2.x, or 5.x) as the nodes in the cluster. Clients may connect to clusters which have a different minor version (e.g. 2.3.x) but it is possible that new functionality may not be supported. Ideally, the client should have the same version as the cluster.

build.gradle
banderous: Good spot @smathangi I got that wrong. I have reinstated the direct dependency, however version 2.10.0 has breaking changes and does not compile. I'd like to keep this PR scoped to the minimum change necessary to resolve our CVEs.

build.gradle
banderous: I have reverted the elasticsearch version bump.

PR #474, opened 2019-11-14
Moving to hmcts version of liquibase which performs better DB locking

adusumillipraveen:
This is based on hmcts/liquibase#2

```
[ ] Yes
[x ] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -219,7 +219,8 @@ subprojects { subproject ->
         testCompileOnly 'org.apiguardian:apiguardian-api:1.0.0'
 
         testCompile group: 'com.github.tomakehurst', name: 'wiremock', version: '2.18.0'
-        testCompile group: 'com.opentable.components', name: 'otj-pg-embedded', version: '0.12.0'
+        testCompile "org.testcontainers:postgresql:1.12.3"
+        testCompile group: 'commons-io', name: 'commons-io', version: '2.6'
         testCompile "org.hamcrest:hamcrest-core:${hamcrestVersion}"
         testCompile "org.hamcrest:hamcrest-library:${hamcrestVersion}"
         testCompile "org.junit.jupiter:junit-jupiter-api:${junitJupiterVersion}"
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -219,7 +219,8 @@ subprojects { subproject ->
         testCompileOnly 'org.apiguardian:apiguardian-api:1.0.0'
 
         testCompile group: 'com.github.tomakehurst', name: 'wiremock', version: '2.18.0'
-        testCompile group: 'com.opentable.components', name: 'otj-pg-embedded', version: '0.12.0'
+        testCompile "org.testcontainers:postgresql:1.12.3"
```
smathangi: Moved to testcontainers from embedded postgres (otj-pg-embedded).
Looks like otj-pg-embedded is not compatible with latest version of liquibase-core as it fails to create sequence for column type "serial" . Similar to the problem mentioned here https://liquibase.jira.com/browse/CORE-1816. I've tried bumping `otj-pg-embedded` to latest but no luck. The error occurs against otj-pg-embedded only.

Eg: Caused by: org.postgresql.util.PSQLException: ERROR: null value in column "id" violates not-null constraint Detail: Failing row contains (null, TEST2, user@hmcts.net, TEST, workBasketDefaultCaseType, workBasketDefaultState, 2019-11-18 23:17:04.184981, UPDATE, rtaashrie9j1otx@example.com).

repository/src/test/java/uk/gov/hmcts/ccd/definition/store/repository/TestConfiguration.java
```diff
@@ -1,36 +1,11 @@
...
-        LOG.info("Starting postgres with port number {}", port);
-
-        final EmbeddedPostgres pg = EmbeddedPostgres
-            .builder()
-            .setPort(randomPort())
-            .start();
-
-        return pg.getPostgresDatabase();
-    }
+    public void emptyConfig() {
```
smathangi: This class can be deleted but that would add another 14 test files into this PR change list. Will do that refactoring in the next PR :)

PR #481, opened 2019-11-20
Fix duplicate query on /findJurisdictions

banderous:
https://tools.hmcts.net/jira/browse/RDM-6603?filter=-3

### Change description ###
This endpoint is frequently called with a list of IDs to retrieve details of specific jurisdictions.

This line would load the requested jurisdictions, then also load every jurisdiction only to throw them away:

`return idsOptional.map(ids -> jurisdictionService.getAll(ids)).orElse(jurisdictionService.getAll());`

Optional.orElse is not lazily evaluated - it is just a java method chain:

```
[ ] Yes
[x] No
```

### DIFF
--- a/rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/CaseDefinitionController.java
+++ b/rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/CaseDefinitionController.java
@@ -91,7 +91,12 @@ public class CaseDefinitionController {
             @ApiParam(value = "list of jurisdiction references") @RequestParam("ids") Optional<List<String>> idsOptional) {
 
         LOG.debug("received find jurisdictions request with ids: {}", idsOptional);
-        return idsOptional.map(ids -> jurisdictionService.getAll(ids)).orElse(jurisdictionService.getAll());
+
+        if (idsOptional.isPresent()) {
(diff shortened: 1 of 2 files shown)


rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/CaseDefinitionController.java
```diff
@@ -91,7 +91,8 @@ public CaseType dataCaseworkerIdAndJurisdictionIdCaseTypeGet(
             @ApiParam(value = "list of jurisdiction references") @RequestParam("ids") Optional<List<String>> idsOptional) {
 
         LOG.debug("received find jurisdictions request with ids: {}", idsOptional);
-        return idsOptional.map(ids -> jurisdictionService.getAll(ids)).orElse(jurisdictionService.getAll());
+
+        return idsOptional.map(ids -> jurisdictionService.getAll(ids)).orElseGet(jurisdictionService::getAll);
```
mario-paniccia: nice one @banderous 💯

mario-paniccia: just to be picky, shall we add a test to avoid regressions?

PR #492, opened 2019-12-13
RDM-6719: Add Banner Feature to CCD UI

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-6719

### Change description ###

Add Banner to CCD UI based on each Jurisdiction configuration

```
[ ] Yes
[ X] No
```


application/src/test/resources/GetCaseTypesResponseForCCD_TestDefinition_V45.json
```diff
@@ -13,6 +13,10 @@
     "description" : "Content for the Test Jurisdiction.",
```
mario-paniccia: would it be possible to properly model the banner in Object Oriented terms with a Banner object?

excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/service/ImportServiceImpl.java
```diff
@@ -228,9 +228,7 @@ public IdamProperties getUserDetails() {
     }
 
     private JurisdictionEntity importJurisdiction(JurisdictionEntity jurisdiction) {
-        return jurisdictionService.get(jurisdiction.getReference()).orElseGet(() -> {
-            jurisdictionService.create(jurisdiction);
-            return jurisdiction;
-        });
+        jurisdictionService.create(jurisdiction);
+        return jurisdiction;
```
MSancaktutar: Couldn't see why we needed to do this.

kiran-yenigala-hmcts: If Jurisdiction already exists in the DB without banner details, and if user tries to import new definition with banner details added into the definition file, then it is not going to save banner information into db, as the previous logic thinks jurisdiction already exists.

application/src/test/resources/GetCaseTypesResponseForCCD_TestDefinition_V45.json
kiran-yenigala-hmcts: We can do that, and need to change database structure as well, need to introduce new table for Banner and reference jurisdiction id in the new table..

mario-paniccia: @kiran-yenigala-hmcts I'd recommend we do that. That we introduce a banner table. I really don't like introducing banner info into the jur table. I would also have a separate Banner worksheet in the spreadsheet dedicated to the banner. Was having a chat with @rafalkalita and he has my same concerns

mario-paniccia: basically I'd like to see ideally 0 changes on the Jur code

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/EntityToResponseDTOMapper.java
```diff
@@ -221,6 +221,8 @@ private EventComplexTypeEntityToCaseEventFieldComplexListMapper() {
     @Mapping(source = "searchAliasFieldEntity.caseType.reference", target = "caseTypeId")
```
mario-paniccia: commenting here because it doesn't allow me to comment on the spreadsheet files. 
I see we have added many new test spreadsheets. In my opinion this should be avoided. The reason being that this increases that maintainability effort of our tests. Because spreadsheet change, we might add new mandatory elements in the future. And when that happens we'll have to update all the test spreadsheets. 
Do we need to add CCD_TestDefinition_V46_RDM-4324.xlsx? Can't we update an existing one?
Also, I see many spreadsheets probably added to test corner case conditions in func tests or integration tests. Do we need to use func tests or int tests for that? We should be able to test all the corner cases with unit tests. To reduce the amount of int and func test to the bare minimun.

application/src/test/resources/GetCaseTypesResponseForCCD_TestDefinition_V45.json
```diff
@@ -13,7 +13,8 @@
...
     "live_from" : "2017-01-01T00:00:00.000+0000",
     "live_until" : null,
-    "case_types" : [ ]
+    "case_types" : [ ],
+    "banners": []
```
mario-paniccia: In CCD all UI bits of definition are not returned as part of the case type definition. Def Store provides separate endpoints to retrieve the UI definitions under the path `/api/display`. See `DisplayApiController`. Banners is a UI definition, so I think we should add retrieval of banners as a dedicated endpoint in `DisplayApiController`.  Data Store could then retrieve the banners for a certain jur by calling this endpoint.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/banner/BannerServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+            .collect(toList());
+    }
+
+    @Override
+    public List<Banner> getAll() {
+        List<BannerEntity> bannerEntities = bannerRepository.findAll();
+        return bannerEntities.stream()
+            .map(dtoMapper::map)
+            .collect(toList());
+    }
```
MSancaktutar: I can't any reasonable use case for this. Could we remove this or throw an UnsupportedOperationException?

kiran-yenigala-hmcts: Removed this method....

application/src/test/resources/GetCaseTypesResponseForCCD_TestDefinition_V45.json
kiran-yenigala-hmcts: Done changes as requested

excel-importer/src/main/java/uk/gov/hmcts/ccd/definition/store/excel/service/ImportServiceImpl.java
kiran-yenigala-hmcts: done changes

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/EntityToResponseDTOMapper.java
kiran-yenigala-hmcts: done changes

PR #520, opened 2020-01-21
removed unwanted configs from ASE infra env files

smathangi:


smathangi:
> is database_max_pool_size not being used?
> is the elastic search not in use either?

these are passed to web_app settings which is not deployed in ASE anymore @nathan-clark

PR #674, opened 2020-07-22
Rdm 8932

A-R05:

https://tools.hmcts.net/jira/browse/RDM-8932

### Change description ###

Changing request parameter to snake case.

```
[ ] Yes
[x] No
```


rest-api/src/main/java/uk/gov/hmcts/ccd/definition/store/rest/endpoint/DisplayApiController.java
```diff
@@ -126,7 +126,7 @@ public WorkBasketResult displayWorkBasketDefinitionIdGet(
         @ApiResponse(code = 200, message = "Unexpected error")
     })
     public SearchCasesResult displaySearchCasesResultIdGet(
-        @ApiParam(value = "Case Type ID", required = true) @PathVariable("id") String id, @RequestParam(value = "usecase", required = false) String usecase) {
+        @ApiParam(value = "Case Type ID", required = true) @PathVariable("id") String id, @RequestParam(value = "use_case", required = false) String usecase) {
```
danlysiak, line 129: Since it is now considered two words:

```suggestion
        @ApiParam(value = "Case Type ID", required = true) @PathVariable("id") String id, @RequestParam(value = "use_case", required = false) String useCase) {
```

PR #683, opened 2020-08-03
RDM-8698 : exact match queries on phone numbers with or without spaces

sumansreenivasan:

https://tools.hmcts.net/jira/browse/RDM-8698

### Change description ###

Exact match queries on phone numbers with or without spaces

```
[ ] Yes
[X] No
```

### DIFF
--- a/charts/ccd-definition-store-api/values.preview.template.yaml
+++ b/charts/ccd-definition-store-api/values.preview.template.yaml
@@ -21,7 +21,7 @@ java:
 
     # enable whenever required and provide host url to match with corresponding data-store-api
     ELASTIC_SEARCH_ENABLED: true
-    ELASTIC_SEARCH_HOST: ccd-data-store-api-pr-601-es-master
+    ELASTIC_SEARCH_HOST: ccd-data-store-api-pr-1043-es-master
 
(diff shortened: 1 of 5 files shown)


elastic-search-support/src/main/resources/casesIndexSettings.json
```diff
@@ -1,21 +1,28 @@
...
+              "type": "custom",
+              "tokenizer": "keyword",
+              "filter": [
+                  "custom_shingle",
+                  "whitespace_remove"
+              ]
+          }
+      },
+      "filter": {
+          "custom_shingle": {
```
danlysiak, line 14: Maybe we can come up with a different name for this, something which is more descriptive of what it actually does?

charts/ccd-definition-store-api/values.preview.template.yaml
```diff
@@ -21,7 +21,7 @@ java:
 
     # enable whenever required and provide host url to match with corresponding data-store-api
     ELASTIC_SEARCH_ENABLED: true
-    ELASTIC_SEARCH_HOST: ccd-data-store-api-pr-601-es-master
+    ELASTIC_SEARCH_HOST: ccd-data-store-api-pr-1043-es-master
```
MSancaktutar, line 24: This line should be changed  back to pr-601 just before merging to develop.

PR #762, opened 2020-10-20
RDM-8898: Removed credentials

rajaram-nagarajan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/RDM-8898

### Change description ###

Eliminate all Credentials in CCD Repos Exposed for Lower Environments

```
[ ] Yes
[x] No
```


rafalkalita:
No need to remove credentials from application.properties. It is used only in local deployment. We use the secrets from the Vault using https://github.com/hmcts/properties-volume-spring-boot-starter.
https://github.com/hmcts/ccd-definition-store-api/blob/master/application/src/main/resources/bootstrap.yaml#L10 will set the DEFINITION_STORE_DB_PASSWORD
https://github.com/hmcts/ccd-definition-store-api/blob/master/charts/ccd-definition-store-api/values.yaml#L23 will set the DEFINITION_STORE_DB_USERNAME

Secrets are defined in the https://github.com/hmcts/ccd-definition-store-api/blob/master/charts/ccd-definition-store-api/values.yaml#L12 and we might want to add DEFINITION_STORE_DB_USERNAME to the secrets.
Interesting is the definition-store-api-POSTGRES-USER is defined in https://github.com/hmcts/ccd-definition-store-api/blob/master/infrastructure/main.tf#L67, but not used ... I think.

rajaram-nagarajan:
> No need to remove credentials from application.properties. It is used only in local deployment. We use the secrets from the Vault using https://github.com/hmcts/properties-volume-spring-boot-starter.
> https://github.com/hmcts/ccd-definition-store-api/blob/master/application/src/main/resources/bootstrap.yaml#L10 will set the DEFINITION_STORE_DB_PASSWORD
> https://github.com/hmcts/ccd-definition-store-api/blob/master/charts/ccd-definition-store-api/values.yaml#L23 will set the DEFINITION_STORE_DB_USERNAME
> 
> Secrets are defined in the https://github.com/hmcts/ccd-definition-store-api/blob/master/charts/ccd-definition-store-api/values.yaml#L12 and we might want to add DEFINITION_STORE_DB_USERNAME to the secrets.
> Interesting is the definition-store-api-POSTGRES-USER is defined in https://github.com/hmcts/ccd-definition-store-api/blob/master/infrastructure/main.tf#L67, but not used ... I think.

Resolved

rafalkalita (approved):
LGTM

PR #852, opened 2020-12-08
RDM-10549 Fix CVEs Newly Reported on All 4 Java Apps

skandala-cgi:
https://tools.hmcts.net/jira/browse/RDM-10549

### Change description ###
Fix CVEs Newly Reported on All 4 Java Apps

```
[ ] Yes
[X] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ dependencyCheck {
     suppressionFile = 'dependency-check-suppressions.xml'
 
     // Ignore these non-runtime configurations
-  //  skipConfigurations = ['codacy']
+   skipConfigurations = ['codacy']
 }
 
 //spring boot dependency override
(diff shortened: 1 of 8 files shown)


MSancaktutar (approved):
Lgtm

timja:
Build failed for: UnknownHostException: ccd-user-profile-api-pr-343.service.core-compute-preview.internal

I've retriggered it

commons/build.gradle
```diff
@@ -16,5 +16,5 @@ bootJar {
 }
 
 dependencies {
-    compile "com.fasterxml.jackson.core:jackson-databind:2.10.3"
+    compile "com.fasterxml.jackson.core:jackson-databind:2.11.1"
```
timja, line 19: not the best way of doing it, you're duplicating the configuration across multiple subprojects.

you should be able to force it at the top level with something like:
https://github.com/hmcts/ccd-user-profile-api/blob/RDM-10549-master/build.gradle#L436

and then not specifying the version anywhere else.

Still approving as it looks like it continues an existing pattern and it's better to have the build green, but would be a nice small improvement imo

build.gradle
```diff
@@ -71,8 +71,8 @@ ext {
     reformLogging= '5.1.1-BETA'
     restAssuredVersion = '4.3.0!!'
     groovyVersion = '3.0.2!!'
-    tomcatVersion = '9.0.38!!'
-    hibernateVersion = '5.4.21.Final!!'
+    tomcatVersion = '9.0.40!!'
```
timja, line 74: what do these !! mean?

timja (approved):

PR #905, opened 2021-01-20
RDM-10879 Set flyway.ignoreMissingMigrations to true.

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-10879

### Change description ###
When changing from develop to master, we have had migrations that are ahead and this was causing the error and preventing the service start. This property disables the check.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/application/src/main/resources/application.properties
+++ b/application/src/main/resources/application.properties
@@ -121,6 +121,6 @@ flyway.noop.strategy=${FLYWAY_NOOP_STRATEGY:false}
 spring.flyway.baseline-on-migrate=true
 spring.flyway.baseline-version=0001
 spring.flyway.out-of-order=true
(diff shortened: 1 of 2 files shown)


application/src/main/resources/application.properties
```diff
@@ -121,5 +121,6 @@ flyway.noop.strategy=${FLYWAY_NOOP_STRATEGY:false}
 spring.flyway.baseline-on-migrate=true
 spring.flyway.baseline-version=0001
 spring.flyway.out-of-order=true
+spring.flyway.ignore-missing-migrations=true
```
smathangi, line 124: you might need to update gradle config as well.

rafalkalita: added

smathangi (approved):

PR #1070, opened 2021-08-20
changed id generation strategy with sequence

kaanaktas:

https://tools.hmcts.net/jira/browse/CCD-1323

### Change description ###
changed id generation strategy with sequence

```
[ ] Yes
[X] No
```


MSancaktutar (approved):
LGTM

repository/src/main/resources/db/migration/V20210820__CCD-1323_change_of_increment_size.sql
```diff
@@ -0,0 +1,6 @@
+ALTER SEQUENCE public.event_post_state_id_seq INCREMENT BY 50;
```
rajaram-nagarajan, line 1: A minor correction may require in the file name - V20210820_1323__CCD-1323_change_of_increment_size.sql

rajaram-nagarajan (approved):
LGTM

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/AccessProfileServiceImpl.java
```diff
@@ -53,7 +53,7 @@ public UserRole getRole(final String role) {
             entity = toEntity(userRole);
             roleFound = false;
         }
-        return new ServiceResponse<>(toModel(repository.save(entity)), roleFound ? UPDATE : CREATE);
+        return new ServiceResponse<>(toModel(repository.saveAndFlush(entity)), roleFound ? UPDATE : CREATE);
```
smathangi, line 56: any particular reason for flush here? ( i didn't see our logic needs to read the saved changes at a later point during the same transaction)

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/AccessProfileEntity.java
```diff
@@ -38,7 +38,7 @@
 public class AccessProfileEntity implements Serializable {
     @Id
     @Column(name = "id")
-    @GeneratedValue(strategy = IDENTITY)
+    @GeneratedValue(strategy = SEQUENCE, generator = "role_id_seq")
```
smathangi, line 41: Just wonder if you have considered `pooled-lo` optimiser already? 
`<property name="hibernate.id.optimizer.pooled.preferred" value="pooled-lo" />`
https://vladmihalcea.com/hibernate-hidden-gem-the-pooled-lo-optimizer/

smathangi (approved):

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/AccessProfileServiceImpl.java
kaanaktas: shouldSaveUserRole() is failing because of createdAt property. As insert is kept in hibernate memory, and createdAt is null at this point and we return the mapped object in the transaction, we are getting nullPointerException for createdAt property. As I understand from the journey, we need to return createdAt to the client, so I directly flushed it and got createdAt's value.

repository/src/main/java/uk/gov/hmcts/ccd/definition/store/repository/entity/AccessProfileEntity.java
kaanaktas: I haven't seen this Sateesh, will read it and try, then compare the performance. I will let you know

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/AccessProfileServiceImpl.java
smathangi: Ok, i think that model mapper should be outside of the Transaction actually - existing tech debt :(

MSancaktutar:
Merged through batch release path.

PR #1080, opened 2021-09-22
implement concurrency pattern to HiddenFieldsValidator

kaanaktas:

https://tools.hmcts.net/jira/browse/CCD-1897

### Change description ###

implement concurrency pattern to HiddenFieldsValidator

```
[ ] Yes
[x] No
```


application/src/main/java/uk/gov/hmcts/ccd/definition/store/ApplicationConfiguration.java
```diff
@@ -36,6 +36,20 @@ public Executor asyncExecutor(@Value("${async.executor.core.pool.size}") Integer
         return executor;
     }
 
+    @Bean("validateExecutor")
+    public Executor validateExecutor(@Value("${validate.executor.core.pool.size:8}") Integer corePoolSize,
+                                     @Value("${validate.executor.max.pool.size:8}") Integer maxPoolSize,
```
MSancaktutar, line 41: better if we combine all default values into app properties - though it may not have been adhered to in few other places.

MSancaktutar (changes requested):
nice logic but it may throw an Exception for a definition item different from the original code. Not sure of the implication of this in terms of impact to service user cases. Perhaps a further touch to assure exception thrown at all times for the same item?

application/src/main/java/uk/gov/hmcts/ccd/definition/store/ApplicationConfiguration.java
kaanaktas: default values were removed, as they are initialized via application.properties under any circumstances.

kaanaktas:
> nice logic but it may throw an Exception for a definition item different from the original code. Not sure of the implication of this in terms of impact to service user cases. Perhaps a further touch to assure exception thrown at all times for the same item?

We now collect all exceptions that occur per validation and return to the client. So they can check and fix the definition files at once.

aat/src/aat/resources/features/F-093/S-600.2.td.json
```diff
@@ -25,7 +25,7 @@
       "Content-Length": "[[ANYTHING_PRESENT]]"
     },
     "body": {
-      "__plainTextValue__" : "'retainHiddenValue' has been incorrectly configured or is invalid for fieldID ['CollectionComplexField'] on ['CaseEventToFields']"
+      "__plainTextValue__" : "'retainHiddenValue' has been incorrectly configured or is invalid for fieldID ['CollectionComplexField'] on ['CaseEventToFields']'retainHiddenValue' has been incorrectly configured or is invalid for fieldID ['CollectionComplexField'] on ['CaseEventToFields']'retainHiddenValue' has been incorrectly configured or is invalid for fieldID ['CollectionComplexField'] on ['CaseEventToFields']'retainHiddenValue' has been incorrectly configured or is invalid for fieldID ['CollectionComplexField'] on ['CaseEventToFields']"
```
MSancaktutar, line 28: Could this please be at least separated by a delimiter like , or \n for better readability?

MSancaktutar: I mean, in all similar places.

kaanaktas: In the api response, I am separating lines with \n delimiter. But,  **__plainTextValue__** is populated after \n delimiter is removed from the result in BeftaFw, and the result comes as I have put in the test case. Please check it out below line in BeftaFw;

https://github.com/hmcts/befta-fw/blob/e815128ada4c703d10f9c9c8e85629481abaca8f/src/main/java/uk/gov/hmcts/befta/player/DefaultBackEndFunctionalTestScenarioPlayer.java#L468

PR #1077, opened 2021-09-15
RDM-12770 Align Global Search field names with existing CCD CaseDetails

padmapriyanalam:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/RDM-12770

### Change description ###
Align Global Search field names with existing CCD CaseDetails

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

PR #1085, opened 2021-10-01
RDM-12686 adjust GlobalSearch index mappings

mattnayler:

   [RDM-12686](https://tools.hmcts.net/jira/browse/RDM-12686) _"Establish the new Global Search endpoint"_

### Change description ###

Adjust `globalSearchCasesMapping.json`:
 * `created_date` required for default sort order for GlobalSearch.
 * `security_classification` adjusted to match default for case indexes
    for use by ElasticsearchSecurityClassificationFilter
 * removed unused `data_classification` from index

**NB**: this PR points to the branch used by #1077.

```
[ ] Yes
[x] No
```

### DIFF
--- a/elastic-search-support/src/main/resources/globalSearchCasesMapping.json
+++ b/elastic-search-support/src/main/resources/globalSearchCasesMapping.json
@@ -1,5 +1,13 @@
 {
   "properties": {
+    "created_date": {
+      "type": "date",
+      "ignore_malformed": true
+    },
+    "last_modified": {
+      "type": "date",
+      "ignore_malformed": true
+    },
     "reference": {
       "type": "text",
       "fields": {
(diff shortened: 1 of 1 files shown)


elastic-search-support/src/main/resources/globalSearchCasesMapping.json
```diff
@@ -232,264 +232,6 @@
         }
       }
     },
-    "data_classification": {
-      "type": "object",
```
smathangi, line 236: Though we don't use this in the response but  i just wonder access control code in ElasticSearchOperation code required `data_classification` ?

mattnayler: It should still be in the response from the ES search, just not available to serach upon.

smathangi: Ok, does ElasticSearch will auto create / update required default analysers based on first record on the fly?  What if we got second case record where structure has new fields (searchParty etc) ?

smathangi: After thinking a while you're @mattnayler we probably don't need any predefined mappings for `data_classification` as they are all String properties and also we never search upon them.

smathangi (approved):

elastic-search-support/src/main/resources/globalSearchCasesMapping.json
mattnayler: As discussed I have re-added `data_classification` but set it as `"enabled": false` to stop it being added by dynamic mapping, see  in b1158d458b13f673282ed1c8dc5ec79905e3470e.

smathangi (approved):

PR #1114, opened 2022-01-04
Rdm 13013

Thor-tech-of-metal:
RDM-13013: Include CaseAccessCategory in the standard search fields for the new ES Global Search Index
           [RDM-13013] (https://tools.hmcts.net/jira/browse/RDM-13013).

### DIFF
--- a/elastic-search-support/src/main/resources/application.yml
+++ b/elastic-search-support/src/main/resources/application.yml
@@ -91,6 +91,7 @@ elasticsearch:
     created_date: ${elasticsearch.elasticMappings.defaultDate}
     last_modified: ${elasticsearch.elasticMappings.defaultDate}
     security_classification: ${elasticsearch.elasticMappings.defaultKeyword}
+    case_access_category:   ${elasticsearch.elasticMappings.defaultText}
     case_type_id: ${elasticsearch.elasticMappings.defaultText}
     last_state_modified_date: ${elasticsearch.elasticMappings.defaultDate}
 #    '[@timestamp]' brackets added for getting along with spring boot 2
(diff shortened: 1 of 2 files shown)


elastic-search-support/src/main/resources/globalSearchCasesMapping.json
```diff
@@ -56,6 +56,18 @@
       "analyzer": "standard",
       "search_analyzer": "alphanumeric_string_analyzer"
     },
+    "case_access_category": {
```
danlysiak, line 59:
```suggestion
    "CaseAccessCategory": {
```
A couple of points:
1. The field id should be `CaseAccessCategory`
1. It is a top level case data field so should be within the `data` object (alongside `SearchCriteria`, `caseManagementLocation` etc.)

elastic-search-support/src/main/resources/application.yml
```diff
@@ -91,6 +91,7 @@ elasticsearch:
     created_date: ${elasticsearch.elasticMappings.defaultDate}
     last_modified: ${elasticsearch.elasticMappings.defaultDate}
     security_classification: ${elasticsearch.elasticMappings.defaultKeyword}
+    case_access_category:   ${elasticsearch.elasticMappings.defaultText}
```
danlysiak, line 94:
```suggestion
```
We should be able to delete this as these properties won't be getting used for the global search index

elastic-search-support/src/main/resources/globalSearchCasesMapping.json
Thor-tech-of-metal: DONE

elastic-search-support/src/main/resources/application.yml
Thor-tech-of-metal: DONE

danlysiak (approved):
🎈

PR #1186, opened 2022-03-18
ccd-2527 clean up, adding delete sqls

duranfatih:

### Change description ###

```
[ ] Yes
[x ] No
```

### DIFF
--- a/repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
+++ b/repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
@@ -151,6 +151,12 @@ BEGIN
   DELETE FROM case_type WHERE id IN
     (SELECT id FROM tmp_case_type_ids);
 
+  DELETE FROM search_criteria WHERE case_type_id IN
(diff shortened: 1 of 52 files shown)


repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
```diff
@@ -145,6 +145,12 @@ BEGIN
   DELETE FROM case_type WHERE id IN
     (SELECT id FROM tmp_case_type_ids);
 
+  DELETE FROM search_criteria WHERE case_type_id IN
+    (SELECT id FROM tmp_case_type_ids);
+
+  DELETE FROM search_party WHERE case_type_id IN
+    (SELECT id FROM tmp_case_type_ids);
+
```
MSancaktutar, line 153: better to put these before the previous delete, as otherwise FK constraints will be violated when they are not dropped during deletion.

duranfatih: done

MSancaktutar (approved):

PR #1192, opened 2022-03-24
cve activities

kaanaktas:

https://tools.hmcts.net/jira/browse/CCD-2840

### Change description ###

```
[ ] Yes
[ ] No
```


MSancaktutar (approved):

dependency-check-suppressions.xml
```diff
@@ -62,4 +62,27 @@
...
+		<cve>CVE-2022-26336</cve>
+	</suppress>
+
+	<suppress until="2022-04-25">
+		<notes><![CDATA[
+   file name: poi-ooxml-schemas-4.1.1.jar
+   ]]></notes>
+		<packageUrl regex="true">^pkg:maven/org\.apache\.poi/poi\-ooxml\-schemas@.*$</packageUrl>
+		<cve>CVE-2022-26336</cve>
+	</suppress>
```
MSancaktutar, line 87: Looks to be same CVE suppressed three times. Needs cleanup later on.

kaanaktas: These are 3 separate dependencies for the same CVE. As I remember they are all needed to be past to dependency check

kaanaktas: Yes, we can do it by changing the regex. Beside this, when the dependency report is generated, it provides an option to get suppression automatically per dependency. I directly used it instead of changing the regex.

PR #1197, opened 2022-03-31
error handling for http 401 status code

kaanaktas:

https://tools.hmcts.net/jira/browse/CCD-2892

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/workbasket/WorkBasketUserDefaultService.java
+++ b/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/workbasket/WorkBasketUserDefaultService.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.ccd.definition.store.domain.service.workbasket;
 
-import org.hibernate.service.spi.ServiceException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpEntity;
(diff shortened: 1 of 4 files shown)


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/workbasket/WorkBasketUserDefaultService.java
```diff
@@ -65,9 +64,7 @@ private void putUserProfiles(final List<WorkBasketUserDefault> workBasketUserDef
 
         } catch (Exception e) {
             LOG.error("Problem updating user profile: {} because of ", workBasketUserDefaults, e);
-            throw new ServiceException(String.format("Problem updating user profile: '%s' because of '%s'",
```
MSancaktutar, line 68: Will we loose this detail from eventual API responses? If so, such loss of detail may slightly harm the API Client's experience.

MSancaktutar (approved):
Minor comment - otherwise good.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/service/workbasket/WorkBasketUserDefaultService.java
kaanaktas: The exception block required only for http call, and downstream status code will be enough, as we are not throwing solid 500 status code anymore.

PR #1238, opened 2022-06-24
CCD-3384  Flags field type does not retain values for hidden sub-field

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CCD-3384

### Change description ###

Flags field type does not retain values for hidden sub-fields

```
[ ] Yes
[x] No
```


RebeccaBaker (changes requested):
can we do a sub select on the field type as we can't guarantee the PK will be the same in each environment. ie something like `UPDATE complex_field SET retain_hidden_value = true
where complex_field_type_id in ((select id from field_type where reference = 'FlagDetails' and version = 1 and jurisdiction_id is null), (select id from field_type where reference = 'Flags' and version = 1 and jurisdiction_id is null));`

markdathornehmcts:
> can we do a sub select on the field type as we can't guarantee the PK will be the same in each environment. ie something like `UPDATE complex_field SET retain_hidden_value = true where complex_field_type_id in ((select id from field_type where reference = 'FlagDetails' and version = 1 and jurisdiction_id is null), (select id from field_type where reference = 'Flags' and version = 1 and jurisdiction_id is null));`

Nice. Done.

noronhaa (approved):

PR #1323, opened 2022-12-07
CCD-4068: Removed deletion queries for table noc_config

dinesh1patel:

https://tools.hmcts.net/jira/browse/CCD-4068

### Change description ###

CCD-4068: Commented out deletion queries for table noc_config as it no longer exists in the db

```
[ ] Yes
[X] No
```

### DIFF
--- a/repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
+++ b/repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
@@ -91,9 +91,6 @@ BEGIN
   DELETE FROM display_group WHERE case_type_id IN
     (SELECT id FROM tmp_case_type_ids);
 
-  --DELETE FROM noc_config WHERE case_type_id IN
-  --    (SELECT id FROM tmp_case_type_ids);
-
   DELETE FROM event_webhook WHERE event_id IN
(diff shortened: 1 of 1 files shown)


repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
```diff
@@ -91,8 +91,8 @@ BEGIN
   DELETE FROM display_group WHERE case_type_id IN
     (SELECT id FROM tmp_case_type_ids);
 
-  DELETE FROM noc_config WHERE case_type_id IN
-      (SELECT id FROM tmp_case_type_ids);
+  --DELETE FROM noc_config WHERE case_type_id IN
```
rajaram-nagarajan, line 94: @dinesh1patel - you can delete these lines instead of commenting it out

rajaram-nagarajan (changes requested):

repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
dinesh1patel: Done

rajaram-nagarajan (approved):
LGTM

PR #1346, opened 2023-02-08
CCD-4327 : Addition of tables in delete scripts

dinesh1patel:

https://tools.hmcts.net/jira/browse/CCD-4327

### Change description ###

CCD-4327 : Addition of tables in delete scripts

```
[ ] Yes
[X] No
```


RebeccaBaker (approved):

repository/src/main/resources/db/useful-queries/Remove all definitions for a case type older than 3 months, but leaving max version.sql
```diff
@@ -82,6 +82,21 @@ BEGIN
     (SELECT id FROM case_field WHERE case_type_id IN
         (SELECT id FROM tmp_case_type_ids)
     );
+  
+  DELETE FROM field_type_list_item WHERE field_type_id  in
+    (SELECT field_type_id FROM case_field WHERE case_type_id IN
+      (SELECT id FROM view__case_type_to_remove)
```
mwallaceHMCTS, line 88: Hi @dinesh1patel - is it intended for this query to remove for any id in the view, rather than the filtered lists like the other additional tables?

PR #1422, opened 2023-11-23
update the label of roleOnCase field to 'Role On Case'

aktaskaan:

https://tools.hmcts.net/jira/browse/CCD-4925

### Change description ###

```
[ ] Yes
[ ] No
```


repository/src/main/resources/db/migration/V20211220_12979__RDM-12979__FlagsFlagDetails.sql
```diff
@@ -97,7 +97,7 @@ and base_field_type_id is null))
 );
 
 insert into complex_field (reference, label, security_classification, field_type_id, complex_field_type_id)
-values ('roleOnCase', 'Flag Type', 'PUBLIC',
+values ('roleOnCase', 'Role On Case', 'PUBLIC',
```
RebeccaBaker, line 100: Changing an existing sql file will cause deployed envs to get an error saying this script has already been run but has been modified since and cant be re run. Can we revert this change and stick with the new script please?

aktaskaan: Thanks Rebecca, it is ready 👍

PR #1543, opened 2025-04-08
CME-238: Elasticsearch Reindexing

SammiChong:
https://tools.hmcts.net/jira/browse/CME-238

### Change description ###

```
[ ] Yes
[ x] No
```


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/event/DefinitionImportedEvent.java
```diff
@@ -1,12 +1,27 @@
...
+    public DefinitionImportedEvent(List<CaseTypeEntity> caseTypes, boolean reindex, boolean deleteOldIndex) {
+        super(caseTypes);
+        this.reindex = reindex;
+        this.deleteOldIndex = deleteOldIndex;
+    }
+
     public DefinitionImportedEvent(List<CaseTypeEntity> caseTypes) {
         super(caseTypes);
+        this.reindex = false;
+        this.deleteOldIndex = true;
```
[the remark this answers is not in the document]
SammiChong: I agree with you, but the ticket mentions the default is true. https://tools.hmcts.net/jira/browse/CME-238 What do you think is the best way to go?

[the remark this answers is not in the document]
danlysiak: Good point, my fault! 😄 I'll update the ticket to have default as `false`.

danlysiak:
More generally I think we could do with a few more tests too 🙏

elastic-search-support/src/test/java/uk/gov/hmcts/ccd/definition/store/elastic/SynchronousElasticDefinitionImportListenerIT.java
```diff
@@ -132,10 +139,60 @@ void shouldCreateElasticsearchIndexForAllCaseTypes() throws IOException, JSONExc
             ignoreFieldsComparator(getDynamicIndexResponseFields(CASE_TYPE_A, CASE_TYPE_B))));
     }
 
+    @Test
+    void shouldReindexSuccessfully() throws JSONException {
```
SammiChong, line 143: Async test issues. Passes in debug but fails when run. I've tried Thread.Sleep() and an await().atMost(50, seconds).untilAsserted which should be enough time but it still fails.

danlysiak: It looks like this fails because the synchronous `elasticClient` is closed almost immediately, so the `reindexData()` call is stopped.. I think to get around this we'll probably need to initialise another new client in  `ElasticDefinitionImportListener::handleReindexing` which we use for for calling `reindexData()`, rather than passing through the one from `initialiseElasticSearch()`. That newly initialised one can then be closed at the same time as the asynchronous one.

kiran-yenigala-hmcts (approved):
LGTM

elastic-search-support/src/main/java/uk/gov/hmcts/ccd/definition/store/elastic/ElasticDefinitionImportListener.java
```diff
@@ -41,40 +48,116 @@ public ElasticDefinitionImportListener(CcdElasticSearchProperties config, CaseMa
...
+                    elasticClient.createIndexAndMapping(incrementedCaseTypeName, caseMapping);
 
-                caseMapping = mappingGenerator.generateMapping(caseType);
-                log.debug("case mapping: {}", caseMapping);
-                elasticClient.upsertMapping(baseIndexName, caseMapping);
+                    //initiate reindexing
+                    handleReindexing(baseIndexName, caseTypeName, incrementedCaseTypeName,
+                        deleteOldIndex);
+                    //dummy value for phase 1
+                    event.setTaskId("taskID");
```
danlysiak, line 88: As agreed for phase 1 ✅ Since the Elastic-provided reindex method doesn't return this and it's not really important for the initial changes since we don't consume it in the frontend anyway, it's not worth making use of the low-level REST client just to get the task ID now; we can revisit alternatives (or removing it) in later phases.

danlysiak (approved):

PR #1557, opened 2025-06-06
CCD-6386/7

patelila:
Add field to the CaseMessage base type to support Query Management thread termination. CCD-6387
Add field to the CaseMessage base type for Query Management message type

- [ ] the `keep-helm` label has been added, if the helm release should be persisted after a successful build

### Change description ###

```
[ ] Yes
[ ] No
```


JamesCollettCGI (approved):

andywilkinshmcts (commented):
How does field optionality work? There's nothing in here that explicitly specifies that these attributes should be optional. Should there be?

patelila:
> How does field optionality work? There's nothing in here that explicitly specifies that these attributes should be optional. Should there be?

I think the field optionality comes from the settings in the Definition file.

PR #1581, opened 2025-07-30
Fix swagger ft

dinesh1patel:

TBA

### Change description ###

Fix Swagger config and uncomment ignore on FT

```
[ ] Yes
[X] No
```


.github/workflows/swagger.yml
dinesh1patel, line 1: Remove this entry '- fix-swagger-ft' prior to merging into master

PR #1750, opened 2026-03-19
CCD-7109 JWT Issuer Validation Disabled in Multiple Services

patelila:

- [ ] the `keep-helm` label has been added, if the helm release should be persisted after a successful build

### Change description ###

```
[ ] Yes
[ ] No
```


dinesh1patel:
closing and repopening to try force Jira re-synch

lang-ben (approved):

application/src/main/java/uk/gov/hmcts/ccd/definition/store/SecurityConfiguration.java
paulridout: Could use constructor injection rather than field injection to initialise the fields annotated with `@Value`. This could make testing the class easier as there wouldn't be a need to use reflection to change the values (if needed).

application/src/main/java/uk/gov/hmcts/ccd/definition/store/SecurityConfiguration.java
```diff
@@ -38,6 +40,9 @@ public class SecurityConfiguration {
     @Value("${oidc.issuer}")
     private String issuerOverride;
```
paulridout: This is now the expected issuer, so it probably shouldn't have "Override" in the name.

application/src/test/java/uk/gov/hmcts/ccd/definition/store/SecurityConfigurationTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.security.oauth2.jwt.JwtClaimValidator;
+import org.springframework.security.oauth2.jwt.JwtTimestampValidator;
+import uk.gov.hmcts.ccd.definition.store.security.OidcIssuerConfiguration;
+
+import java.time.Instant;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class SecurityConfigurationTest {
```
paulridout: This class is testing an instance of `OAuth2TokenValidator<Jwt>` created by it's `validator()` method.  Although the validator is created in the same way as it is on the SecurityConfiguration class, it's not testing the actual one created by the SecurityConfiguration class.  If the SecurityConfiguration class is changed then these tests become invalid.  The class needs to be changed to test something that's produced by SecurityConfiguration.

application/src/test/java/uk/gov/hmcts/ccd/definition/store/security/JwtDecoderExceptionTest.java
```diff
@@ -0,0 +1,121 @@
...
+import uk.gov.hmcts.net.ccd.definition.store.util.KeyGenerator;
+
+import java.text.ParseException;
+import java.time.Instant;
+import java.util.Date;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class JwtDecoderExceptionTest {
```
paulridout: This class is testing an instance of `JwtDecoder` created by it's `jwtDecoder()` method. Although the decoder is created in the same way as it is on the SecurityConfiguration class, it's not testing the actual one created by the SecurityConfiguration class. If the SecurityConfiguration class is changed then these tests become invalid. The class needs to be changed to test something that's produced by SecurityConfiguration.

commons/src/test/java/uk/gov/hmcts/ccd/definition/store/security/OidcIssuerConfigurationTest.java
paulridout, line 1: Might want to add tests where allowed issuer list contains nothing before or after a comma, e.g. ", secondary", "secondary," and "secondary,,tertiary".

commons/src/test/java/uk/gov/hmcts/ccd/definition/store/security/OidcIssuerConfigurationTest.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Test
+    void shouldIncludePrimaryAndConfiguredAllowedIssuers() {
+        assertThat(OidcIssuerConfiguration.allowedIssuers("primary", " secondary, tertiary , secondary "))
+            .containsExactly("primary", "secondary", "tertiary");
+    }
+
+    @Test
+    void shouldRejectBlankPrimaryIssuerEvenWhenAllowedIssuersAreConfigured() {
```
paulridout: Could make this a JUnit5 parameterised test to increase the test coverage, e.g.

```
@ParameterizedTest
@ValueSource(strings = {" "})
@NullAndEmptySource
void shouldRejectMissingPrimaryIssuer(String primaryIssuer) {
    assertThatThrownBy(() -> OidcIssuerConfiguration.allowedIssuers(primaryIssuer, "secondary"))
            .isInstanceOf(IllegalStateException.class)
            .hasMessage("oidc.issuer must not be blank");
}
```

commons/src/test/java/uk/gov/hmcts/ccd/definition/store/security/OidcIssuerConfigurationTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class OidcIssuerConfigurationTest {
+
+    @Test
+    void shouldFallbackToPrimaryIssuerWhenAllowedIssuersUnset() {
```
paulridout: "Fallback" doesn't seem the right term to use when referring to the primary issuer. Perhaps change "FallbackTo" to "OnlyHave".

This also applies to the `shouldFallbackToPrimaryIssuerWhenAllowedIssuersBlank()` test name.

charts/ccd-definition-store-api/values.yaml
```diff
@@ -72,5 +72,7 @@ java:
     IDAM_USER_URL: https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net
     IDAM_S2S_URL: http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
     USER_PROFILE_HOST: http://ccd-user-profile-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
-    OIDC_ISSUER: https://forgerock-am.service.core-compute-idam-{{ .Values.global.environment }}.internal:8443/openam/oauth2/hmcts
+    # Enforced JWT issuer. This must match the exact iss claim in live IDAM access tokens.
+    # Do not infer it from the discovery URL; verify it from a real token for the target environment.
+    OIDC_ISSUER: https://forgerock-am.service.core-compute-idam-aat2.internal:8443/openam/oauth2/realms/root/realms/hmcts
```
paulridout, line 77: Seems a bit odd to set the prime issuer to this value as it's the legacy issuer and will need to be overridden in all environments other than AAT. By way of comparison, the JWT issuer validation changes for aac-manage-case-assignment set it to the `idam-web-public` value.

May also need to add `OIDC_ALLOWED_ISSUERS` environment variable to allow hmcts-access and forgerock issuers.

aat/src/aat/java/uk/gov/hmcts/ccd/definitionstore/befta/JwtIssuerVerificationApp.java
```diff
@@ -0,0 +1,48 @@
...
+        String actualIssuer = resolveIssuerFromRealToken(AATHelper.INSTANCE);
+
+        if (!allowedIssuers.contains(actualIssuer)) {
+            throw new IllegalStateException(
+                "OIDC issuer mismatch: expected one of `" + String.join("`, `", allowedIssuers)
+                    + "` but token iss was `" + actualIssuer + "`"
+            );
+        }
+
+        System.out.println("Verified functional test token iss is allowed: " + actualIssuer);
```
paulridout: Can SLF4J logging be used here instead of `System.out.println()`?

charts/ccd-definition-store-api/values.yaml
patelila: OIDC_ALLOWED_ISSUERS is optional and only takes effect when explicitly set. If unset/blank, issuer validation falls back to OIDC_ISSUER only. If set, it is additive: the validator accepts OIDC_ISSUER plus each exact comma-separated issuer in OIDC_ALLOWED_ISSUERS.
I agree we should avoid guessing issuer values. The current value was set to match the issuer seen by the AAT smoke/functional test token path. If this service also needs to accept hmcts-access/public IDAM tokens, then we should verify the real token iss values and configure the primary issuer plus OIDC_ALLOWED_ISSUERS accordingly, mirrored in Jenkins for the verifier. I would avoid changing the Helm default to idam-web-public without that token evidence.

commons/src/test/java/uk/gov/hmcts/ccd/definition/store/security/OidcIssuerConfigurationTest.java
```diff
@@ -0,0 +1,75 @@
...
+            .hasMessage("oidc.issuer must not be blank");
+    }
+
+    private static Stream<Arguments> allowedIssuerListsWithEmptyEntries() {
+        return Stream.of(
+            Arguments.of(
+                "nothing before comma",
+                ", " + SECONDARY_ISSUER,
+                new String[]{PRIMARY_ISSUER, SECONDARY_ISSUER}
+            ),
```
paulridout, line 61: The first argument is only used to provide a name for the test.  To avoid needing to add a corresponding parameter to the test method which isn't used by it, the JUnit `named()` method can be used to provide a custom name for the first argument which is then used as the test name.  For example:

```
import static org.junit.jupiter.api.Named.named;

...

@ParameterizedTest(name = "{index}: {0}")
@MethodSource("allowedIssuerListsWithEmptyEntries")
void shouldIgnoreEmptyConfiguredAllowedIssuerEntries(String configuredAllowedIssuers, String[] expectedIssuers) {
   ...
}

private static Stream<Arguments> allowedIssuerListsWithEmptyEntries() {
    return Stream.of(
        Arguments.of(
            named("nothing before comma", ", " + SECONDARY_ISSUER),
            new String[]{PRIMARY_ISSUER, SECONDARY_ISSUER}
        ),
        ...
    );
}
```

The `{index}` placeholder in the parameterised test name just displays the number of the test (1, 2, etc.).  It can be left out if desired.

paulridout (approved):
Changes look OK.

hemantt (approved):

PR #1796, opened 2026-05-01
CCD-7525 Fix both parsing issues - the not equals parsing and the clo…

markdathornehmcts:

https://tools.hmcts.net/jira/browse/CCD-7525

### Change description ###

Fix FieldShow condition issues.
1. when != is in the condition
2. when combined parentheses are used, the closing parenthesis can be lost  

```
[ ] Yes
[x] No
```

### DIFF
--- a/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/showcondition/ShowConditionParser.java
+++ b/domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/showcondition/ShowConditionParser.java
@@ -29,7 +29,7 @@ public class ShowConditionParser {
         "\\s*?(.*)\\s*?(=|CONTAINS)\\s*?(\".*\"\\)*)\\s*?");
     @SuppressWarnings("squid:S5852")
     private static final Pattern NOT_EQUAL_CONDITION_PATTERN = Pattern.compile(
-        "\\s*?(.*)\\s*?(!=|CONTAINS)\\s*?(\".*\")\\s*?");
+        "\\s*?(.*)\\s*?(!=|CONTAINS)\\s*?(\".*\"\\)*)\\s*?");
     @SuppressWarnings("squid:S5852")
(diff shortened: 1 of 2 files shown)


domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/showcondition/ShowConditionParser.java
```diff
@@ -29,7 +29,7 @@ public class ShowConditionParser {
         "\\s*?(.*)\\s*?(=|CONTAINS)\\s*?(\".*\"\\)*)\\s*?");
     @SuppressWarnings("squid:S5852")
     private static final Pattern NOT_EQUAL_CONDITION_PATTERN = Pattern.compile(
```
paulridout, line 31: To keep in with the naming convention used by the equality constants, we might want to now rename this to NOT_EQUAL_CONDITION_PATTERN_WITH_TRAILING_BRACKET.

domain/src/main/java/uk/gov/hmcts/ccd/definition/store/domain/showcondition/ShowConditionParser.java
```diff
@@ -41,6 +41,7 @@ public class ShowConditionParser {
     private static final Pattern INCORRECT_POSITION_PATTERN = Pattern.compile(INCORRECT_POSITION_OF_PARENTHESIS_REGEX);
     private static final Pattern INCORRECT_POSITION_CONDITION_PATTERN =
         Pattern.compile(INCORRECT_POSITION_OF_CONDITION_REGEX);
+    private static final Pattern QUOTED_VALUE_PATTERN = Pattern.compile("\"[^\"]*\"");
```
paulridout, line 44: As an alternative, you could use lazy matching instead of `[^\"]*` in this regular expression.  The whole regular expression would then be `\".*?\"`.  As with the current regular expression, this will only match the first double quote found after the initial one.

